PR #11, opened 2022-08-24
Descriptors component

will-byrne:
Add the descriptors component with the doi, open access, and license link.


src/components/atoms/descriptors/descriptors.tsx
```diff
@@ -0,0 +1,23 @@
...
+    </ul>
+    <ul className="descriptors__icons">
+      <li>
+        <a href="https://en.wikipedia.org/wiki/Open_access" className="descriptors__icon descriptors__icon--oa">
+          <span className="visuallyhidden">Open access</span>
+        </a>
+      </li>
+      <li>
+        <a href="https://creativecommons.org/licenses/by/4.0/" className="descriptors__icon descriptors__icon--cc">
+          <span className="visuallyhidden">Copyright information</span>
```
nlisgo, line 18: Will all reviewed preprints by published under the same license? Of course, we can determine that this is the case for the initial examples.

nlisgo (approved):

src/components/atoms/descriptors/descriptors.tsx
will-byrne: This would be a question for the production team i think, we modelled this on the version we had before. It may be an enhancement that we worry about after Oct (or at least when we know we need it).

PR #12, opened 2022-08-24
add title component

will-byrne:


src/sass/_typography.scss
```diff
@@ -31,3 +31,11 @@
   font-size: 14px;
   line-height: 24px;
 }
+
+@mixin heading-1() {
```
discodavey, line 35: Not a fan of calling heading-1 when h1 will be matching what it is doing. Just my opinion

will-byrne: done

scottaubrey:
I think this will need the ability to inject HTML into the title, or something more fancy like the ability to parse some formatting within the component, right? Some of our example preprints utilise this already, but I'm pretty certain it's a base level expectation that the title can involve some formatting.

will-byrne:
> 

this is doable but we probably want to come up with a list of allowed html elements and decide what shape the data passed to the component will look like.

src/sass/_typography.scss
```diff
@@ -31,3 +31,11 @@
   font-size: 14px;
   line-height: 24px;
 }
+
+@mixin h1() {
+  color: settings.$color-text;
+  font-family: settings.$font-family-primary;
+  font-size: 36px;
```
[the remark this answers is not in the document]
discodavey: They are. We will do this as another ticket. Probably Tuesday. We will do a blanket refactor

nlisgo:
It would be amazing if we could apply filters to parameters. Here is a library that we could configure to allow only the permitted html into the template: https://www.npmjs.com/package/sanitize-html

will-byrne:
> It would be amazing if we could apply filters to parameters. Here is a library that we could configure to allow only the permitted html into the template: https://www.npmjs.com/package/sanitize-html

It is considered bad practice to just put raw html into a component and would probably be better to construct it from what we parse out of the xml than to render it as html and and put it in that way (even if its sanitised).

scottaubrey:
I was imagining something similar to [this on the server](https://github.com/elifesciences/enhanced-preprints-server/blob/master/src/model/content.ts#L38) - a function that transforms a format with encoded and limited text transformations to HTML equivalent. I'm not necessarily suggesting that we use the encoda-formatted JSON as there (though it is ready, and type-defined), but a similar structure with similar purpose that can encoded enough of the JATS standard formatting we'd expect would work also. This transform function could then be used by different react components for rendering the limited-format content we feed it across different parts of the UI, but taking just "text" for others where we don't expect or want formatting.

This would then need to feed back into the server code that transforms the XML to store in the DB so that we only transform and store the format-encoded output, and we transmit that to the client via the API.

From a chat with @will-byrne this is similar to what he was thinking - but we both think it's worth further discussion and prototyping together to understand the problem space better.

nlisgo:
> > It would be amazing if we could apply filters to parameters. Here is a library that we could configure to allow only the permitted html into the template: https://www.npmjs.com/package/sanitize-html
> 
> It is considered bad practice to just put raw html into a component and would probably be better to construct it from what we parse out of the xml than to render it as html and and put it in that way (even if its sanitised).

That would be a server app concern presumably.

src/components/atoms/title/title.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  it('should render the title with a single content part passed into the title prop', () => {
+    const title = { type: 'Strong', content: 'This is a title' };
+    render(<Title title={title}/>);
+
+    expect(screen.getByText('This is a title')).toBeInTheDocument();
+  });
+
+  it('should render the title with an array of content parts passed into the title prop', () => { // whitespace issues
```
scottaubrey, line 26: This is exactly how encoda's output works at the moment. Which sort of makes sense - the array should only be used to decorate parts of a piece of content (e.g. Emphasis inside a paragraph). encoda seems to include a space in the string before and after, and this produces correct output. 

Given we're talking about HTML output though - whitespace is relatively insignificant, and it probably doesn't matter if you want to add extra space between array elements?

src/components/atoms/title/title.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const title = [{ type: 'Strong', content: 'This is a' }, { type: 'Emphasis', content: 'title' }];
+    render(<Title title={title}/>);
+
+    expect(screen.getByText('This is a')).toBeInTheDocument();
+    expect(screen.getByText('title')).toBeInTheDocument();
+  });
+
+  it('should render the title with an array of content parts and strings passed into the title prop', () => { // components in a list need a key
```
scottaubrey, line 34: @will-byrne Not sure I understand what this means? Is it a React limitation?

will-byrne: yes anything in react that is an output of a loop needs a key so react can link the dom / shadow dome elements

discodavey (approved):

PR #19, opened 2022-09-02
add article-content component and contentToJsx utility

will-byrne:
adding the ArticleContent component.
We will need to decide whether we want a separate component for each content type and where we will put styling.

### DIFF
--- a/src/utils/content-to-jsx.test.tsx
+++ b/src/utils/content-to-jsx.test.tsx
@@ -1,0 +1,131 @@
+import { contentToJsx } from './content-to-jsx';
+import { Heading } from '../components/atoms/heading/heading';
+
+describe('Content to JSX', () => {
+  it('returns the string unchanged if passed a simple string', () => {
+    const result = contentToJsx('foo');
+
+    expect(result).toStrictEqual('foo');
+    expect(typeof result).toBe('string');
+  });
+
+  it('returns an array of jsx components and strings if passed an array', () => {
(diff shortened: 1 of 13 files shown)


src/utils/content-to-jsx.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+        <figcaption>I am a caption</figcaption>
+      </figure>,
+    );
+  });
+
+  it('generates the expected html when passed a ImageObject', () => {
+    const result = contentToJsx({
+      type: 'ImageObject',
+      contentUrl: 'https://placekitten.com/500/300',
+      content: [],
```
scottaubrey, line 123: I'm guessing you've added this to satisfy the type. Would it be better to alter `ImageObjectContent` so it's not a union with `DecoratedContent` ?

scottaubrey: Actually, following the [Stencila Schema definition for `ImageObject`](https://github.com/stencila/schema/blob/master/ts/types.ts#L1088), I think only contentUrl is required, content can be there, but should be optional.

The type definitions in the link above are actually quite helpful in understanding encoda's JSON output.

scottaubrey (approved):
I've listed some small improvements, but I'm mostly personally happy for this to merge as is. 

I think if you agree with any of the comments, we should create tickets and do them as follow up work, rather than avoid merging this PR.

PR #33, opened 2022-09-07
add jump to menu

scottaubrey:
- add jump to menu
- add test
- lint fixes

### DIFF
--- a/src/components/atoms/jump-to-menu/jump-to-menu.scss
+++ b/src/components/atoms/jump-to-menu/jump-to-menu.scss
@@ -1,5 +1,5 @@
-@use '../../../sass/settings';
-@use '../../../sass/typography';
+@use "../../../sass/settings";
+@use "../../../sass/typography";
 
 .jump-menu-container {
   border-left: 2px solid settings.$color-less-light-grey;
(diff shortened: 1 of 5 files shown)


will-byrne (approved):

src/components/atoms/jump-to-menu/jump-to-menu.scss
```diff
@@ -0,0 +1,47 @@
...
+  padding-left: 20px;
+}
+
+.jump-menu-list__link {
+  @include typography.navigation-heading();
+
+  text-decoration: none;
+}
+
+.jump-menu-list__item--active .jump-menu-list__link {
```
discodavey, line 45: Can we reduce the specificity on this one?

scottaubrey: Do you want to move the `--active` flag down to the link? We wouldn't be able to just remove the parent class as the logic currently is.

discodavey: Ok. Well leave it as it is and I will look at it when I review all of the CSS

PR #40, opened 2022-09-09
add tabbed navigation

scottaubrey:
- fix just about working Tabbed navigation
- add test and styles
- lint fixes

### DIFF
--- a/src/components/molecules/tabbed-navigation/tabbed-navigation.tsx
+++ b/src/components/molecules/tabbed-navigation/tabbed-navigation.tsx
@@ -20,9 +20,7 @@ export const TabbedNavigation = ({ children, initiallySelected = 0 }: TabbedNavi
           return (<li className={`tabbed-navigation__tab-label ${activeTab === index ? 'tabbed-navigation__tab-label--active' : ''}`} key={label} onClick={() => setActiveTab(index)}>{label}</li>);
         })}
       </ul>
-      <div className="tabbed-navigation__content">
-        {children[activeTab]}
-      </div>
+      {children[activeTab]}
     </div>
   );
 };
--- a/src/components/molecules/tabbed-navigation/tabbed-navigation.tsx
+++ b/src/components/molecules/tabbed-navigation/tabbed-navigation.tsx
@@ -1,12 +1,15 @@
 import { ReactElement, useState } from 'react';
 import { TabProps } from './tab';
 
+import './tabbed-navigation.scss';
+
 type TabbedNavigationProps = {
+  initiallySelected?: number,
   children: ReactElement<TabProps>[],
 };
 
-export const TabbedNavigation = ({ children }: TabbedNavigationProps): JSX.Element => {
-  const [activeTab, setActiveTab] = useState(0);
+export const TabbedNavigation = ({ children, initiallySelected = 0 }: TabbedNavigationProps): JSX.Element => {
+  const [activeTab, setActiveTab] = useState(initiallySelected);
 
   return (
     <div className="tabbed-navigation">
@@ -14,12 +17,7 @@ export const TabbedNavigation = ({ children }: TabbedNavigationProps): JSX.Eleme
         {children.map((child, index) => {
           const { label } = child.props;
 
-          return (
-            <li
-              key={label}
-              onClick={() => setActiveTab(index)}
-            >{label}</li>
-          );
+          return (<li className={`tabbed-navigation__tab-label ${activeTab === index ? 'tabbed-navigation__tab-label--active' : ''}`} key={label} onClick={() => setActiveTab(index)}>{label}</li>);
         })}
       </ul>
       <div className="tabbed-navigation__content">
--- a/src/components/molecules/tabbed-navigation/tabbed-navigation.tsx
+++ b/src/components/molecules/tabbed-navigation/tabbed-navigation.tsx
@@ -1,0 +1,30 @@
+import { ReactElement, useState } from 'react';
+import { TabProps } from './tab';
+
+type TabbedNavigationProps = {
+  children: ReactElement<TabProps>[],
+};
+
+export const TabbedNavigation = ({ children }: TabbedNavigationProps): JSX.Element => {
+  const [activeTab, setActiveTab] = useState(0);
+
+  return (
+    <div className="tabbed-navigation">
+      <ul className="tabbed-navigation__tabs">
+        {children.map((child, index) => {
+          const { label } = child.props;
+
+          return (
+            <li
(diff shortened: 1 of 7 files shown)


src/components/molecules/tabbed-navigation/tabbed-navigation.tsx
```diff
@@ -0,0 +1,28 @@
...
+  initiallySelected?: number,
+  children: ReactElement<TabProps>[],
+};
+
+export const TabbedNavigation = ({ children, initiallySelected = 0 }: TabbedNavigationProps): JSX.Element => {
+  const [activeTab, setActiveTab] = useState(initiallySelected);
+
+  return (
+    <div className="tabbed-navigation">
+      <ul className="tabbed-navigation__tabs">
```
discodavey, line 16: Will need to aria labels etc to this for screen readers

src/components/molecules/tabbed-navigation/tabbed-navigation.tsx
```diff
@@ -0,0 +1,28 @@
...
+        {children.map((child, index) => {
+          const { label } = child.props;
+
+          return (<li className={`tabbed-navigation__tab-label ${activeTab === index ? 'tabbed-navigation__tab-label--active' : ''}`} key={label} onClick={() => setActiveTab(index)}>{label}</li>);
```
discodavey, line 20: Will need to aria labels etc to this for screen readers

discodavey (commented):
I know some of this is dummy content but need to make sure there are no divs or spans that are empty and to make sure they have classes in them.

Screenshot 2022-09-09 at 08 38 47

src/components/molecules/tabbed-navigation/tabbed-navigation.tsx
```diff
@@ -0,0 +1,28 @@
...
+        })}
+      </ul>
+      <div className="tabbed-navigation__content">
```
discodavey, line 23: Possible aria labels etc needed here

discodavey (commented):
Content is not sat in the div and overlaps. This just seems to be the image. Will need to add some dummy content to make sure this works ok as well.

Screenshot 2022-09-09 at 08 44 22

Just added text and this looks ok. Just images at the moment that need to be looked at.

Screenshot 2022-09-09 at 08 51 08

scottaubrey:
> I know some of this is dummy content but need to make sure there are no divs or spans that are empty and to make sure they have classes in them.
> 
> Screenshot 2022-09-09 at 08 38 47

Now looks like this:
Screenshot 2022-09-09 at 9 46 15 am

Is that OK @discodavey ?

scottaubrey:
> Content is not sat in the div and overlaps. This just seems to be the image. Will need to add some dummy content to make sure this works ok as well.
> 
> Screenshot 2022-09-09 at 08 44 22
> 
> Just added text and this looks ok. Just images at the moment that need to be looked at.
> 
> Screenshot 2022-09-09 at 08 51 08

I think the container has no width. If we change the width of the image (via the URL params) it overlaps more/less. 

I assume this is a wider styling issue, but are you want the tab content to enforce containing the internal content? If so, in what way, `overflow: hidden/scroll`? or a min-width with sub-styles? Sorry, I'm not sure exactly what you're looking for?

src/components/molecules/tabbed-navigation/tabbed-navigation.tsx
scottaubrey: Never used them. Something we can pair on together?

discodavey:
> > I know some of this is dummy content but need to make sure there are no divs or spans that are empty and to make sure they have classes in them.
> > Screenshot 2022-09-09 at 08 38 47
> 
> Now looks like this: Screenshot 2022-09-09 at 9 46 15 am
> 
> Is that OK @discodavey ?

Yes. Just need to get rid of the empty div's. Always been one of my hates for React.

discodavey:
We just need to sort the images out in the content section

scottaubrey:
> We just need to sort the images out in the content section

@discodavey See my comment above - how are you expecting that to happen inside the tab? Should we be specifying Tab with `overflow: ` behaviour specified? Given we don't know what will be in it (or where it might be), I'm reluctant to add anything too specific to accommodate this random cat pic, and most "content" will actually be pulled in through the ArticleContent component which will have and need styling to contain random content, and feel more appropriate place to be specifying image width/contraints

will-byrne:
> > We just need to sort the images out in the content section
> 
> @discodavey See my comment above - how are you expecting that to happen inside the tab? Should we be specifying Tab with `overflow: ` behaviour specified? Given we don't know what will be in it (or where it might be), I'm reluctant to add anything too specific to accommodate this random cat pic, and most "content" will actually be pulled in through the ArticleContent component which will have and need styling to contain random content, and feel more appropriate place to be specifying image width/contraints

We might want to pass a prop of className to the tab component and proxy it to the outer element of the container. That should let us easily change the root class of the generated content. As it will probably be different for article content and reviews.

src/components/molecules/tabbed-navigation/tabbed-navigation.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <div className="tabbed-navigation">
+      <ul className="tabbed-navigation__tabs">
+        {children.map((child, index) => {
+          const { label } = child.props;
+
+          return (<li className={`tabbed-navigation__tab-label ${activeTab === index ? 'tabbed-navigation__tab-label--active' : ''}`} key={label} onClick={() => setActiveTab(index)}>{label}</li>);
+        })}
+      </ul>
+      {children[activeTab]}
```
will-byrne, line 23: This should probably be in a container like it was before so we can add things like padding when the content is added. It also makes the content siblings of the tab headers which feels wrong and could have issues with layout.

scottaubrey: Each <Tab> element creates a `<div class=tabbed-navigation_content />` with it's content within. Children here is a collection of those divs. 

So I think it does do what you're asking, just a level below as part of the <Tab> element that is showing. I think it's probably doable to have <Tab> just splurge it's children one way or another though, if it's easier to read/preferable?

will-byrne: Thats true, i missed that. 😁

discodavey (approved):

scottaubrey:
Spoke with @discodavey he's happy that content inside a tab should define it's expected content behaviour (like `ArticleContent`). @will-byrne are you OK with that - at least for now?

PR #69, opened 2022-09-29
Fix storybook for Article Page

### LINKED ISSUE #68: ArticlePage story is not showing in storybook
# Description

Loading up storybook doesn't load the story from `src/pages/[...doi]/index.stories.tsx` and doesn't load in storybook.

scottaubrey:
This PR moves the bulk of the page setup back into a standalone React Component, including stories and SASS. This fixes the story not turning up in storybook.

I've then created a shell of what remains to setup nextjs page, and import the Page React Component above.

I've tested locally, but would be good for @discodavey to give it a go and see if it resolves for him too before merging.

Fixes #68


src/components/pages/article/article-page.page.tsx
```diff
@@ -0,0 +1,52 @@
+import { ArticleContent } from '../../atoms/article-content/article-content';
```
will-byrne, line 1: this file doesnt need to be `.page.tsx` anymore

will-byrne, line 1: this file doesnt need to be `.page.tsx` anymore

will-byrne (changes requested):

src/components/pages/article/article-page.page.tsx
scottaubrey: Correct!

will-byrne (approved):

PR #73, opened 2022-09-30
use image server

scottaubrey:
- add image server and basic figure component
- add basic styles to figure component
- encode the URI component when passing into the image server URL
- fix data import to use publicly-available articles
- fix issue with imageserver not showing when javascript was on


scottaubrey:
In particular, I'm not sure if [this](https://github.com/elifesciences/enhanced-preprints-client/pull/73/commits/367ca423a2a26699c69eef1ccc01b5be9d467cdf) is the correct way to do this, but it feels iffy returning a custom component from a function called `contentToJsx()`

scottaubrey:
> In particular, I'm not sure if [this](https://github.com/elifesciences/enhanced-preprints-client/pull/73/commits/367ca423a2a26699c69eef1ccc01b5be9d467cdf) is the correct way to do this, but it feels iffy returning a custom component from a function called `contentToJsx()`

Any thoughts on this @will-byrne - are we OK with this?

will-byrne:
> > In particular, I'm not sure if [this](https://github.com/elifesciences/enhanced-preprints-client/pull/73/commits/367ca423a2a26699c69eef1ccc01b5be9d467cdf) is the correct way to do this, but it feels iffy returning a custom component from a function called `contentToJsx()`
> 
> Any thoughts on this @will-byrne - are we OK with this?

Its returning a jsx component so i dont have a problem with it, its really no different than using a third party library that returns a component.

PR #74, opened 2022-09-30
fix: add anchor IDs for figures (and tables)

scottaubrey:


discodavey:
ID's will need to be different for each figure and table.

scottaubrey:
> ID's will need to be different for each figure and table.

Yup, they are different in the content source.

will-byrne (approved):

Issue #56, opened 2022-09-26
Review components


scottaubrey:
# Description

We need to reimplement the reviews tab components in EPP client

# Tasks

- [ ] Editors list
- [ ] Reviews HTML

![Image](https://user-images.githubusercontent.com/375812/193536321-bcc44a8a-4d1a-42d2-912c-981fc26fff14.png)

scottaubrey:
Work completed in multiple pull requests.

Prerequisites:
https://github.com/elifesciences/enhanced-preprints-client/pull/76 - Adds page published via manuscript ID, by mapping locally which msids are preprints.

Pull Requests:
https://github.com/elifesciences/enhanced-preprints-server/pull/391 - Fixes some missing data on institutions required for the Editors view
https://github.com/elifesciences/enhanced-preprints-client/pull/79 - Adds an atom for Editors list
https://github.com/elifesciences/enhanced-preprints-client/pull/80 - Adds an atom for Review component, built with `dangerouslySetInnerHTML` and `marked`
https://github.com/elifesciences/enhanced-preprints-client/pull/81 - Queries the API in the server PR, and passes it to the atoms above in the Peer Review tab

PR #80, opened 2022-10-03
Review components review content

scottaubrey:
References https://github.com/elifesciences/enhanced-preprints-client/issues/56


src/components/atoms/review-content/review-content.tsx
```diff
@@ -0,0 +1,6 @@
+import { marked } from 'marked';
+import styles from './review-content.module.scss';
+
+export const ReviewContent = ({ content }: { content: string }): JSX.Element => (
+  <section className={styles['review-content']} dangerouslySetInnerHTML={{ __html: marked.parse(content) }}></section>
```
will-byrne, line 5: Isnt the review content we get from the docmap already html?

scottaubrey: Correct! You passed the test! Your special prize is working with a forgetful idiot, Congratulations! 🤦 

Not sure why I had forgotten since I've been saying for a while we don't need marked anymore on the server. I just straight up lifted it as an example from server, and didn't question bringing in marked at all.

Anyway, pushed a simpler version now.

will-byrne (approved):

PR #111, opened 2022-10-11
Fix routing for journal proxying

scottaubrey:
To allow clean proxy behaviour, we tell nextjs to publish all content at `/reviewed-preprints`. We then change the main page route to be from the basePath so that /reviewed-preprints/12345 still resolves. 

~This change does mean `/` can no longer route from epp client and will result in a 404 or 5xx on the EPP system url (i.e. not through journal).~

I came up with a better way that redirects wayward requests to the correct place, and allows us to prefix assets in prod to the journal Url. 

This whole journal integration will need a look at again after launch, but this will at least work in a non-journal specific way.

### DIFF
--- a/.storybook/fix-next-image.js
+++ b/.storybook/fix-next-image.js
@@ -1,0 +1,13 @@
+import * as NextImage from "next/image";
+
+const OriginalNextImage = NextImage.default;
+
+Object.defineProperty(NextImage, "default", {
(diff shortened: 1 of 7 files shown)


scottaubrey:
Sorry I know this has been through a few additional commits since I ask for review, but it's now ready for review @will-byrne

PR #113, opened 2022-10-12
Remove curation labels

discodavey:


scottaubrey (changes requested):
Looking good. 

I think given we've removed the isMSA styling, we probably want to remove `isMSA` arg and the class (that I think results in an `undefined` class on the msas) from `article-flag.tsx`, related tests and stories, `article-flag-list.tsx` and related stories and tests.

After that it looks good to me

scottaubrey (approved):

PR #115, opened 2022-10-12
Add main navigation

### LINKED ISSUE #50: Add menu(s) to match new menu on journal
Ensure menus are aligned across both Journal and EPP. Next step to reduce duplication of dev across two platforms for single changes

will-byrne:
Closes #50


scottaubrey (approved):
The SaSS is quite difficult to determine which bits are used, and we would like to rethink this later on any. Chris I think has ideas to reduce bits of it too. We also don't currently had any tests.

So on the basis that we'll review multiple times soon anyway, and it does at least look like I think it should look, ALL THE THUMBS UP.

Issue #103, opened 2022-10-10
Add /status endpoint


scottaubrey:
# Description

To aid service heath checks and integration into the journal, we should create an endpoint at /status that checks all dependant services are available and can connect, and  returns with 200 status on success, or 5xx on failure

Dependant services to check:
- client -> server
- server -> database server
- server -> IIIF server
- client -> IIIF server ?

Q: Do we need a status on both server and client? Does client call server?

# Definition of Done

- [ ] Can hit https://staging--epp.elifesciences.org/status (or https://prod--epp.elifesciences.org/status) and receive a HTTP status code 200 or 5xx if and only if services are failing.

Issue #102, opened 2022-10-10
Add /ping endpoint


scottaubrey:
# Description

To aid service heath checks and integration into the journal, we should create an endpoint at /ping that simply returns `pong` with 200 status to check the web server is responding.

# Definition of done 
- [ ] `curl https://staging--epp.elifesciences.org/ping` returns status 200 and "pong" as a body

PR #121, opened 2022-10-12
Status API

scottaubrey:
Doesn't work. *points at nextjs*

It will fix https://github.com/elifesciences/enhanced-preprints-client/issues/102 and fix https://github.com/elifesciences/enhanced-preprints-client/issues/103


scottaubrey:
This now works. It's an API with a `.page` suffix 🤷

will-byrne (approved):

PR #136, opened 2022-10-14
refactor so text for search and alerts shows all the time above 1000px

discodavey:


discodavey:
@scottaubrey Don't review as yet

Issue #104, opened 2022-10-10
Google Tag manager / Analytics on EPP client


scottaubrey:
Related: https://github.com/elifesciences/enhanced-preprints-server/issues/9

The EPP client does not have google analytics enabled. 

Definition of Done

- [x] GTM is appearing in the EPP HTML
- [x] Data for EPP staging and prod is appearing in GA
- [x] Hotjar can record a session

de-code:
Manual installation instructons:

> Below is the Google tag for this account. Copy and paste it in the code of every page of your website, immediately after the `<head>` element. Don’t add more than one Google tag to each page.

```html
<!-- Google tag (gtag.js) -->
<script async src="https://www.googletagmanager.com/gtag/js?id=G-CTCHX4ZQV5"></script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());

  gtag('config', 'G-CTCHX4ZQV5');
</script>
```

de-code:
Note: updated the above comment

scottaubrey:
@BlueReZZ @de-code This should be live for both environments now.

BlueReZZ:
Excellent, thank you.

@de-code is going to work on ensuring it is measuring correctly then look at specific measures detailed in https://github.com/elifesciences/data-hub-issues/issues/525

scottaubrey:
Sure - https://staging--epp.elifesciences.org/ and https://prod--epp.elifesciences.org/

de-code:
> Hi @scottaubrey could you please remind me of the URLs for staging and production?

Paul shared the links with me.

de-code:
For Cookiebot (not sure if you want to create it as a separate ticket), it seems to be just adding the following line as the **very first script** element:

```html
<script id="Cookiebot" src="https://consent.cookiebot.com/uc.js" data-cbid="0a5c50d8-fcf9-47b1-8f4f-1eaadb13941b" type="text/javascript"></script>
```

In this case staging and production seem to be using the same id (`0a5c50d8-fcf9-47b1-8f4f-1eaadb13941b`).


For more context, this is from the [Cookiebot Help](https://www.cookiebot.com/en/help/):

[...]

PR #140, opened 2022-10-14
Add in support got GoogleTagManager with injected ID

scottaubrey:
related https://github.com/elifesciences/enhanced-preprints-client/issues/104

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -6,4 +6,5 @@ if (!process.env.API_SERVER && process.env.NEXT_PHASE !== 'phase-production-buil
 export const config = {
   apiServer: process.env.API_SERVER,
   imageServer: process.env.NEXT_PUBLIC_IMAGE_SERVER,
+  gtmId: process.env.NEXT_PUBLIC_GTM_ID,
 };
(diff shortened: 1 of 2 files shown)


will-byrne (approved):

discodavey:
Has the ID for the journal been added as this does not need it's own reporting as it can be done as part of the main journal.

scottaubrey:
> Has the ID for the journal been added as this does not need it's own reporting as it can be done as part of the main journal.
> 

I believe so, see https://github.com/elifesciences/enhanced-preprints-client/issues/104

scottaubrey:
Though again we will need to make and consider what an EPP for biophysics looks like - I'm not sure we want analytics mixed with our own

scottaubrey:
> Paul said last week that we did not need to set up another account for it. Best speak to him

Yeah, for launch that's definitely true. For eLife ongoing that's almost definitely still true. If I had thought otherwise, I would have objected at the time.

I'm was just saying, like other things we've done in EPP for eLife - this might need reconsidering for biophysics.

PR #143, opened 2022-10-17
Jump to menu functionality

will-byrne:
Closes https://github.com/elifesciences/enhanced-preprints-server/issues/423


scottaubrey (changes requested):
There's a weird jumping behaviour where a menu item near the bottom is clicked, it highlights it, then as the page scrolls it jumps back up and highlights menu items on the way down.

I wonder if it would be better to add a timeout for onClick to `setActive`, but bypass `handleScroll` affect? Another var maybe that just checks if we're jumping, with a timeout of the scroll time?

PR #148, opened 2022-10-18
Main content refactor (figures)

discodavey:


src/components/pages/article/article-page.tsx
```diff
@@ -108,14 +108,32 @@ export const ArticlePage = (props: { metaData: ArticlePageProps, abstract: Conte
           </div>
         </Tab>
         <Tab label="Figures and data">
-          <Heading id="figures" headingLevel={2} content="Figures and data" />
-          <ArticleContent content={getFigures(props.content)} />
+          <JumpToMenu headings={[
```
scottaubrey, line 111: The JumpMenu in the page component need to be built with the headings from the content before we merge this.

will-byrne:
This doesnt run for me, if you do a `docker-compose up` in the client it complains that:
```
./src/components/atoms/descriptors/descriptors.module.scss:16:1
Syntax error: Selector "img" is not pure (pure selectors must contain at least one local class or id)

  14 | }
  15 | 
> 16 | img {
     |     ^
  17 |   max-height: 100%;
  18 |   max-width: 100%;
```

scottaubrey (changes requested):

discodavey:
> 

This passed all linting and it should accept this

scottaubrey:
Sorry Dave, I'm still getting the issue @will-byrne mentioned
Screenshot 2022-10-18 at 12 14 30 pm

discodavey:
I'll take it out but not liking that we are being told we cannot have this as it is more of an opinion.

There will be other selectors that will need to be put in after Thursday so will need to allow them in the stylesheets

scottaubrey:
Same deal with bare figure
Screenshot 2022-10-18 at 2 28 11 pm

For reference, this isn't linting failures. The style can't be built for production - it's a hard failure. And it's a failure because the selector is part of a component, but isn't scoped to the component. Like a selector lacking BEM-style selector in a sass file alongside a component, just enforced by the build-stage because it has no name to attach it to.

scottaubrey (approved):

PR #157, opened 2022-10-18
Get some browsers to do a better job with rendering weight 600

scottaubrey:
by including weight 700 when loading the font

### DIFF
--- a/src/pages/_document.page.tsx
+++ b/src/pages/_document.page.tsx
@@ -10,7 +10,7 @@ export default function Document() {
   return (
     <Html>
       <Head>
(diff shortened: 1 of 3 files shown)


src/pages/_document.page.tsx
```diff
@@ -10,7 +10,7 @@ export default function Document() {
   return (
     <Html>
       <Head>
-        <link href="https://fonts.googleapis.com/css2?family=Noto+Sans:wght@400;600&family=Noto+Serif:wght@400;600" rel="stylesheet"/>
+        <link href="https://fonts.googleapis.com/css2?family=Noto+Sans:wght@400;600&family=Noto+Serif:wght@400;700" rel="stylesheet"/>
         <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet"/>
```
nlisgo, line 14: This looks good to me.

Is there any reason why we would not use an include for the link fonts to avoid duplication?

nlisgo (approved):

PR #215, opened 2022-11-01
Add download, share, citation and follow icons for buttons

discodavey:


discodavey:
I have coded the buttons like this but there maybe a better way of doing this. @will-byrne or @scottaubrey ?

scottaubrey (approved):

PR #223, opened 2022-11-02
Remove empty divs and refactor tabbed navigation

discodavey:
Opinions welcome on this please?


src/components/molecules/tabbed-navigation/tabbed-navigation.module.scss
```diff
@@ -2,6 +2,10 @@
 @use "../../../sass/settings";
 @use "../../../sass/typography";
 
+.tabbed-navigation {
+  display: block;
+}
+
```
will-byrne, line 8: is this just here so that it shows the class in the dom?

discodavey: Yes. I think we need to have a nav tag around the tabbed navigation

will-byrne (approved):

PR #238, opened 2022-11-04
Make tabs work with no JS

scottaubrey:
This changes the page entirely to define tab content outside of the tab. This allows us to then define 3 different routes for the tabs to display, and changes the tabs to links between them. Due to using React and Next, this then updates the page client side, or navigates to the other route server side. Also due to NextJS and caching - this is very fast either way.

fixes https://github.com/elifesciences/enhanced-preprints-client/issues/240


nlisgo:
I will look at this today!

src/components/pages/article/article-page.test.tsx
```diff
@@ -1,11 +1,55 @@
...
+      </ArticlePage>)).not.toThrow();
+  });
+
+  it('renders with figures tab', () => {
+    expect(() => render(<ArticlePage metaData={metaData} status={status} activeTab="figures">
+        <ArticleFiguresTab content={content} />
+      </ArticlePage>)).not.toThrow();
+  });
+
+  it('renders with figures tab', () => {
```
nlisgo, line 21: is this supposed to say `renders with reviews tab`?

scottaubrey: What would you say if I said no? :P

I've pushed the correction

nlisgo:
The tabs work as expected with and without javascript. I did find that it can be quite slow switching between tabs but if there is no reason to believe it is any different than the current approach then we shouldn't hold up and optimise. We could do that as a followup.

I would argue that if the javascript switching isn't quick we should just have it work without javascript. The experience for users if you click on something and it doesn't react in a timely manner can be jolting.

I found that when javascript was turned off that none of the styling came in. This may be an issue with my local setup. Can you verify that this isn't the experience for you?

scottaubrey:
> The tabs work as expected with and without javascript. I did find that it can be quite slow switching between tabs but if there is no reason to believe it is any different than the current approach then we shouldn't hold up and optimise. We could do that as a followup.
> 
> I would argue that if the javascript switching isn't quick we should just have it work without javascript. The experience for users if you click on something and it doesn't react in a timely manner can be jolting.
> 
> I found that when javascript was turned off that none of the styling came in. This may be an issue with my local setup. Can you verify that this isn't the experience for you?

Both issues are more products of the Dev server setup. If you run the prod mode (`make start-prod`), both should be resolved

nlisgo:
> Both issues are more products of the Dev server setup. If you run the prod mode (make start-prod), both should be resolved

I can confirm that it works much better on both counts when running `make start-prod`.

However, there is one behaviour that is different with javascript compared with the current implementation. When you click on a tab, the correct content loads and the url changes but the page is scrolled to the top. This is not how it works currently. If this is the new expected behaviour then bonus but thought it was worth pointing out if it was unintended.

nlisgo (approved):

scottaubrey:
> However, there is one behaviour that is different with javascript compared with the current implementation. When you click on a tab, the correct content loads and the url changes but the page is scrolled to the top. This is not how it works currently. If this is the new expected behaviour then bonus but thought it was worth pointing out if it was unintended.

Fixed - if you wouldn't mind another look and approval?

nlisgo (approved):
:ship:

PR #242, opened 2022-11-08
Add 404 page

discodavey:


discodavey:
@scottaubrey Is there anyway of being able to get a preview of this so Chris can see it before it goes through staging and into production?

scottaubrey:
@discodavey EPP-prod automations are paused at the moment, so you can merge and see it in staging without it going "live" to prod.

There's no automatic preview instances at the moment. I can work on it, but we need this repo to build the PRs, and we need the cluster to generate the deployment. I've did this as a 10% time a few months back before we had -client, so I know it's all possible. The only issue is the time to put it all together again.

So we're left with manual, which requires manual building the image, then generating a deployment manually. All doable, but it all takes time, and possibly isn't worth it compared to just arranging to share your screen, or just merging to staging and iterating any improvements from there.

src/components/pages/article/article-page.module.scss
```diff
@@ -1,6 +1,7 @@
 @use "../../../sass/settings";
 @use "../../../sass/layout";
 @use "../../../sass/mixins";
+@use "../../../sass/typography";
```
scottaubrey, line 4: This isn't required?

scottaubrey (changes requested):
It works absolutely fine in the app, which is good.

However, I'm getting an error in storybook:
```
Image with src "static/media/404.c2442964a87f3533faee771fc005b486.svg" must use "width" and "height" properties or "layout='fill'" property.
```

Obviously, this isn't stopping a build, but it stopping viewing it on story book. It fills the screen if I use layout=fill (filling the screen), so setting width and height is the only we to get it to go away. When I explicitly set the height/width (`width="64" height="157"`), it does show in sb. 

The storybook view also doesn't have the DefaultLayout in that story (just import and wrap it at the story level I reckon).

Other than that thought it looks good to me.

discodavey:
@scottaubrey I have done the changes from the review. In Storybook it is just showing the content and not the whole page (i.e. missing siteheader). Does this matter?

I think we should have the siteheader as it is a page

scottaubrey:
> @scottaubrey I have done the changes from the review. In Storybook it is just showing the content and not the whole page (i.e. missing siteheader). Does this matter?
> 
> I think we should have the siteheader as it is a page

Yup, agreed. Take a look at the article-page storybook, see if the way that wraps the component helps.

scottaubrey (approved):

PR #246, opened 2022-11-10
Add a preview image build when preview tag is added

scottaubrey:


.github/workflows/build-preview.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Build and load preview image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          target: prod
+          load: true
+          tags: |
+            ${{ env.IMAGE_REPO }}:preview-${{ github.sha }}
```
nlisgo, line 34: the [github.sha](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) is supposed to correspond to the commit sha that triggered the workflow. Firstly, I couldn't confirm that that value did in fact correspond with the commit sha. I see the following build but can not find a commit sha with the same value: https://github.com/elifesciences/enhanced-preprints-client/pkgs/container/enhanced-preprints-client/49935512?tag=preview-f17e6a7c96027975de86f79a2587adba1da0b2ca

This approach, unless I'm mistaken, will prepare and deposit a new image for each commit on a PR with the preview label. This could amount to a lot of deposited builds. Do these builds have a lifetime or should we organise clean up?

scottaubrey: `github.sha` will refer to the last commit that is in the pr merge branch - i.e. it is a commit specifically to show the potential merge commit if merged. You can get to it via special branch refs: `refs/pull/:prNumber/merge`. It's right to build from that SHA though, as it is the most accurate portrayal of the code after merge to the target branch.

As for creating lots of images, yes that's true, but I'm not sure it matters - each master branch image is labelled quite differently, and :latest always gets updated too. Space wise, the layers architecture of docker means most of the layers are identical, and published only once (except where deps are updated, say, and then if the PR gets merged that layer can get reused anyway). 

If you're worried though, the easiest solution would be to publish to separate repo for previews. This can then be periodically purged of all labels whenever we have no preview PRs in play. This would disrupt the layers behaviour above slightly, but I'm not sure I'm bothered much in either direction.

nlisgo (approved):
This will be a valuable feature.

PR #254, opened 2022-11-11
Ensure all institutions load without javascript

nlisgo:


nlisgo:
I took the same approach as we are using for [authors](https://github.com/elifesciences/enhanced-preprints-client/blob/master/src/components/atoms/authors/authors.tsx).

scottaubrey (approved):
LGTM

PR #256, opened 2022-11-16
Make manuscripts config injectable from json file

scottaubrey:
This will allow the cluster to configure this data differently in different environments

refs https://github.com/elifesciences/enhanced-preprints-issues/issues/100


src/config.ts
```diff
@@ -9,4 +9,5 @@ export const config = {
   gtmId: process.env.NEXT_PUBLIC_GTM_ID,
   cookiebotId: process.env.NEXT_PUBLIC_COOKIEBOT_ID,
   articleCacheAge: process.env.ARTICLE_CACHE_AGE || '1800',
+  manuscriptConfigFile: process.env.MANUSCRIPT_CONFIG_FILE || './manuscripts.json',
```
nlisgo, line 12: Do we have a followup task to remove the production json file from this repo and to introduce it from another source to production, staging and feature branch instances. 

We should also consider creating a json schema for the manuscripts json.

scottaubrey: The intention is to remove the manuscripts config file altogether. However, to more fully answer your question - the follow up tasks for this work do include injecting the config in from the cluster, and at that point we _could_ remove it from this repo. 

However, we still pull down production data, and import it as part of the docker-compose setup, and so we'd need to maintain local-only copies of this config for testing purposes in the short term anyway. We could create a schema, or an example file with only some of it in, but the value for a feature we ultimately want to code out of existence seems low to me, what do you think?

PR #257, opened 2022-11-16
Add footer

discodavey:


src/sass/_mixins.scss
```diff
@@ -82,6 +82,18 @@
   }
 }
 
+@mixin height($size_in_px) {
+  height: #{$size_in_px}px;
+  height: #{typography.get-rem-from-px($size_in_px)}rem;
+}
+
+@mixin constrain-height($size_in_px, $extreme) {
```
scottaubrey, line 90: I can't find a use of this, was it copied across from journal by mistake or did you bring it just for symmetry?

src/components/layouts/default.tsx
```diff
@@ -1,11 +1,16 @@
...
+  <>
+    <div className={`${styles['grid-container']} ${styles['article-page']}`}>
+      <div className={styles['grid-header']}>
+        <SiteHeader />
+      </div>
+        {children}
     </div>
-    {children}
-  </div>
+    <Footer/>
```
scottaubrey, line 13: I wonder if `Footer` should be named `SiteFooter` to match `SiteHeader`?

src/components/atoms/investors/investors.module.scss
```diff
@@ -0,0 +1,59 @@
...
+    align-items: center;
+    display: flex;
+    justify-content: center;
+  }
+}
+
+.investor-logos__img {
+  display: block;
+  margin: auto;
+  max-height: $image-height;
```
scottaubrey, line 37: These are showing as invalid property values - need 'px'?

src/components/atoms/investors/investors.module.scss
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.investor-logos__item {
+  @include mixins.padding(24);
+
+  border-top: 1px solid settings.$divider;
+}
+
+.investor-logos__container {
+  @include mixins.constrain-width($image-width, "max");
```
scottaubrey, line 21: I know this was already broken, but this also results in an invalid property value as there are no units for the non-rem value in the mixin

src/components/atoms/footer-main/footer-main.tsx
```diff
@@ -0,0 +1,142 @@
+import Image from 'next/image';
+import carbonNeutral from '../../../images/logos/carbon-neutral.svg';
+import styles from './footer-main.module.scss';
+
+/* eslint-disable max-len */
+
+export const FooterMain = (): JSX.Element => (
+  <footer className={styles['site-footer']}>
+
+    <div className={styles['site-footer__container']}>
```
scottaubrey, line 10: Should this be separate from `<footer>`? Does that make it (or the `<div>`) an empty element? I think it might be necessary for the top-border/divider, but I'm just sanity checking it?

src/sass/_mixins.scss
discodavey: I did bring it over and started to use it but decided it was only being used once so will remove it

src/components/layouts/default.tsx
discodavey: Agree. Will change it to SiteFooter

src/components/atoms/investors/investors.module.scss
discodavey: Done

src/components/atoms/footer-main/footer-main.tsx
discodavey: Footer deals with the border, padding and 100% width where as the container sets up the flex box and sets the max-width to 1296px so both are needed.

scottaubrey (approved):

scottaubrey (approved):

PR #260, opened 2022-11-18
fix the size of funder images 

scottaubrey:
The sizes set in the next/image object are for producing optimized images for mobile and HiDPI screens. This needs to be the target height for display in page - what we want it to be.

This PR sets the next/image component width/height at the appropriate ratio adjusted width in ratio to the height set by other logos. This fixes issues with oversized images rendering when there is a CSS failure.


discodavey (approved):
Looks ok to me.

PR #263, opened 2022-11-22
Allow optional tags to wrap content in contentToString

nlisgo:
This will be useful for reviewed preprints api: https://github.com/elifesciences/enhanced-preprints-issues/issues/1

### DIFF
--- a/src/utils/content-to-string.test.tsx
+++ b/src/utils/content-to-string.test.tsx
@@ -106,31 +106,29 @@ describe('Content to String', () => {
     expect(result).toStrictEqual('');
   });
 
-  it('optionally wraps content in tags', () => {
+  it.each([
     [
-      {
-        content: ['one', 'two', { type: 'Strong', content: 'three' }],
-        contentTypeTags: [{ id: ContentType.strong, tag: 'b' }],
-        expected: 'onetwo<b>three</b>',
-      },
-      {
-        content: ['one', 'two', { type: 'Strong', content: 'three' }],
-        contentTypeTags: [{ id: ContentType.strong, tag: 'strong' }],
-        expected: 'onetwo<strong>three</strong>',
-      },
-      {
-        content: ['one', 'two', { type: 'Paragraph', content: ['three', { type: 'Subscript', content: 'four' }, 'five'] }],
-        contentTypeTags: [{ id: ContentType.strong, tag: 'b' }, { id: ContentType.subscript, tag: 'sub' }],
-        expected: 'onetwothree<sub>four</sub>five',
-      },
-      {
-        content: ['one', 'two', { type: 'Paragraph', content: ['three', { type: 'Subscript', content: 'four' }, 'five'] }],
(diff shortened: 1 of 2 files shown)


nlisgo:
This commit on the reviewed preprints api PR shows what benefit this will bring: https://github.com/elifesciences/enhanced-preprints-client/pull/245/commits/b41e499f7f3e28a1cfd0ccc6d74d59188391875c

nlisgo:
@will-byrne I am using jest-each

scottaubrey:
I have no issues specifically with the implementation or tests per se, but I think we're overloading a little too much on the one function, and I'm not sure this can honestly be called `contentToString` anymore, rather it's `contentToSubsetOfHtml`. 

However, I don't really like the idea of creating a third `contentTo*` function, but I'm struggling for a better suggestion than that. Maybe @will-byrne might have a suggestion

nlisgo:
> I have no issues specifically with the implementation or tests per se, but I think we're overloading a little too much on the one function, and I'm not sure this can honestly be called `contentToString` anymore, rather it's `contentToSubsetOfHtml`.
> 
> However, I don't really like the idea of creating a third `contentTo*` function, but I'm struggling for a better suggestion than that. Maybe @will-byrne might have a suggestion

If a new function was preferred we could share some code I'm sure. I justified the change because it is still a string (a string of HTML) :)

scottaubrey:
> > I have no issues specifically with the implementation or tests per se, but I think we're overloading a little too much on the one function, and I'm not sure this can honestly be called `contentToString` anymore, rather it's `contentToSubsetOfHtml`.
> > However, I don't really like the idea of creating a third `contentTo*` function, but I'm struggling for a better suggestion than that. Maybe @will-byrne might have a suggestion
> 
> If a new function was preferred we could share some code I'm sure. I justified the change because it is still a string (a string of HTML) :)

Yeah, I thought of that after I finished last night. In retrospect, my intention when I created it was `contentToPlainText` and naming is hard.

I wonder if there is a way to reuse contentToJsx, i.e. a wrapper around it that renders to HTML string?

scottaubrey:
> > > I have no issues specifically with the implementation or tests per se, but I think we're overloading a little too much on the one function, and I'm not sure this can honestly be called `contentToString` anymore, rather it's `contentToSubsetOfHtml`.
> > > However, I don't really like the idea of creating a third `contentTo*` function, but I'm struggling for a better suggestion than that. Maybe @will-byrne might have a suggestion
> > 
> > 
> > If a new function was preferred we could share some code I'm sure. I justified the change because it is still a string (a string of HTML) :)
> 
> Yeah, I thought of that after I finished last night. In retrospect, my intention when I created it was `contentToPlainText` and naming is hard.
> 
> I wonder if there is a way to reuse contentToJsx, i.e. a wrapper around it that renders to HTML string?

@nlisgo like [this](https://github.com/elifesciences/enhanced-preprints-client/compare/content-to-html?expand=1) maybe?

nlisgo:
Thanks @scottaubrey! I'm closing this PR in favour of: https://github.com/elifesciences/enhanced-preprints-client/pull/271

PR #267, opened 2022-11-24
fix issue with placement of next/title prerendering in the wrong place

scottaubrey:
```
Warning: <title> should not be used in _document.js's <Head>. https://nextjs.org/docs/messages/no-document-title
```


nlisgo (approved):
:+1:

PR #273, opened 2022-11-25
Fix tabbed navigation

discodavey:


scottaubrey (approved):
Checked-out, it's sticky again!

PR #277, opened 2022-11-30
Reset the body margin to 0

nlisgo:


scottaubrey (approved):

src/pages/_document.page.tsx
```diff
@@ -18,6 +18,7 @@ export default function Document() {
           }
           body {
             background-color: #fff;
```
discodavey, line 20: This does not need to be added as the default is white. We will need to have future conversations about changing the background to a slight off white as it's not good for people with dyslexia.

nlisgo: > We will need to have future conversations about changing the background to a slight off white as it's not good for people with dyslexia.

Create a ticket for it.

PR #278, opened 2022-12-05
Add modal window

discodavey:


src/components/atoms/modal/modal.tsx
```diff
@@ -0,0 +1,29 @@
...
+  };
+
+  return (
+  <>
+    <button className="modal-button" onClick={() => setShowModal(!showModal)}>Modal Link</button>
+    <div onClick={(event) => clickHandler(event)} className={`${styles['modal-container']} ${showModal ? styles['modal-content__show'] : ''} `}>
+      <div ref={contentRef} className={styles['modal-content']}>
+        <button className={styles['modal-content__close-button']} onClick={closeModal}>Close</button>
+        <div className="modal-content__block">
+          <h6>Modal Window Title</h6>
```
scottaubrey, line 22: Can we make these two a prop before merging?

src/components/atoms/modal/modal.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  const clickHandler = (event: MouseEvent<HTMLDivElement>) => {
+    if (contentRef.current && !contentRef.current.contains(event.target as Element)) {
+      closeModal();
+    }
+  };
+
+  return (
+  <>
+    <button className="modal-button" onClick={() => setShowModal(!showModal)}>Modal Link</button>
```
scottaubrey, line 17: I know why this is here, but we probably should make this a prop to start with, and demo how to trigger with a button - unless you think just getting it in for now is valu.... a worthwhile isolated step to deliver ;)

scottaubrey (approved):

PR #280, opened 2022-12-05
Hide institutions and display author line on narrow viewports

nlisgo:


discodavey:
@nlisgo Code looks ok. If you want to go through it just in case I have missed something then just shout.

Probably best to do this tomorrow morning

src/components/atoms/authors/authors.module.scss
```diff
@@ -18,19 +19,97 @@
...
+      @media only all and (max-width: settings.$breakpoint-medium) {
+        &.authors-list__expansion-count-3 {
+          display: inline;
+        }
+      }
 
-  cursor: pointer;
-  font-weight: 600;
+      @media only all and (min-width: settings.$breakpoint-medium) {
+        &.authors-list__expansion-count-10 {
```
discodavey, line 97: Just gone through this again. Why do we have magic numbers on classes like here and on line 91, 104 and 110?

nlisgo:
@chris-huggins can you check that it is working as expected with and without JavaScript:

https://pr-280--epp.elifesciences.org/

src/components/atoms/authors/authors.tsx
```diff
@@ -2,25 +2,33 @@ import { useEffect, useState } from 'react';
...
           <li className={styles['authors-list__item']} key={index}>
             {givenNames.join(' ')} {familyNames.join(' ')}
           </li>
         ))}
       </ol>
-      {(authors.length > authorLimit && expanded !== null) && <span className={styles['authors-list__expansion']} onClick={() => setExpanded(!expanded)}>{expansionText}</span>}
+        {(authors.length > authorLimit && expanded !== null) &&
+        <span className={styles['authors-list__expansion']} onClick={() => setExpanded(!expanded)}>
+          show{!expanded ? authorLimits.map(
+          (limit, index) => <span key={index} aria-hidden="true" className={`${styles['authors-list__expansion-count']} ${styles[`authors-list__expansion-count-${limit}`]}`}> {authors.length - limit}</span>,
```
nlisgo, line 30: After a chat with @discodavey we have added `aria-hidden`. The inclusion of the counts for screen readers will be confusing but the text without the count will still make sense.

nlisgo:
@scottaubrey is it possible to re-deploy the demo instance from the latest commit.

src/components/atoms/authors/authors.module.scss
```diff
@@ -18,19 +19,94 @@
...
+        display: none;
+      }
+    }
+
+    &:last-of-type { /* stylelint-disable-line no-duplicate-selectors */
+      display: inline;
+    }
+  }
+
+  .authors-list--expanded & { /* stylelint-disable-line no-descending-specificity */
```
nlisgo, line 60: @discodavey I would appreciate your input to see if these lint disable can be avoided. I don't want to assume that I can't refactor this and conform with the linting and have the intended effect.

nlisgo:
> @scottaubrey is it possible to re-deploy the demo instance from the latest commit.

@chris-huggins this is ready for your feedback again. I have implemented some of the feedback that you have given including:

- display final author on all view port sizes preceded by the ellipsis if required at that viewport size. 
- amend the weight of the expansion trigger from 600 to 400

src/components/atoms/authors/authors.module.scss
discodavey: What is the linting error?

nlisgo: > What is the linting error?

`no-descending-specificity`

nlisgo:
@chris-huggins has reviewed the behaviour in the demo instance. Just need a review from a member of @elifesciences/epp-tech-team

scottaubrey (approved):

scottaubrey:
Should we create a ticket to better test the hiding behaviour at the browser level?

nlisgo:
> Should we create a ticket to better test the hiding behaviour at the browser level?

I would love to introduce tests like that but understand that it is not trivial.

nlisgo:
Followup ticket to improve js tests: https://github.com/elifesciences/enhanced-preprints-issues/issues/25#issuecomment-1354700505

PR #291, opened 2022-12-16
Fix unspecified typing in test suite

scottaubrey:


nlisgo:
@scottaubrey is this issue caused by a different type of author than one that is anticipated? In which case this could unblock us but we need to capture the requirement of what authorLine to present in that scenario instead of just stepping over that author.

scottaubrey:
The type error is caused by this expectation value:
https://github.com/elifesciences/enhanced-preprints-client/blob/173d0682a6e1c9a4ee5badfa594c4f755f3ca106/src/pages/api/reviewed-preprints.page.test.ts#L150-L153

but the "issue" is caused by this block of code:
https://github.com/elifesciences/enhanced-preprints-client/blob/173d0682a6e1c9a4ee5badfa594c4f755f3ca106/src/pages/api/reviewed-preprints.page.ts#L41-L44

Which in tern lands, and then is given a new (but identical) value here:
https://github.com/elifesciences/enhanced-preprints-client/blob/173d0682a6e1c9a4ee5badfa594c4f755f3ca106/src/pages/api/reviewed-preprints.page.ts#L104

You could change the original return to a `return '';`, change the return value of the function, then change the intermediate function call to not be a ternary, then change the test expectation, and the typing change in this PR wouldn't be required. I guess though that you were looking to return no value when it is changed to JSON (which undefined causes the JSON to not contain the key). In which case, the undefined is the correct expectation, and the type change above is a correct representation of the expected inputs from those examples.

nlisgo:
The intention behind `return;` was to return undefined for authorLine if there were no authors. The undefined properties seemed to be filtered out in the response without the need to filter. Prior to understanding that I may have sought to only set a property of `authorLine` if I knew it would have a value. If it is blank then I don't want it in the response body.

scottaubrey:
> The intention behind `return;` was to return undefined for authorLine if there were no authors. The undefined properties seemed to be filtered out in the response without the need to filter. Prior to understanding that I may have sought to only set a property of `authorLine` if I knew it would have a value. If it is blank then I don't want it in the response body.

Exactly as I suspected. So the page function is returning the javascript object with undefined. When that's encoded to JSON, it's not set as a key. But when we test what the page function returns, it needs to be an `undefined`. As so we need this extra type to be tested for. And so we need that extra type in the test function expectatation.

nlisgo (approved):

PR #293, opened 2022-12-19
Disabled modular css

scottaubrey:
Fixes https://github.com/elifesciences/enhanced-preprints-issues/issues/369


nlisgo (approved):
Looks good to me. Does this mean that we can add global styles?

scottaubrey:
> Looks good to me. Does this mean that we can add global styles?

We always could add global styles, we just never got around to doing it right. Thanks for the review!

PR #294, opened 2022-12-19
update gha outputs to use the newer redirect-based output

scottaubrey:
relates to https://github.com/elifesciences/enhanced-preprints-issues/issues/324


nlisgo (approved):
I think we have the same issue on server?

PR #306, opened 2023-01-05
Introduce form-item text-field into component library

nlisgo:


src/components/atoms/forms/form-item.scss
```diff
@@ -0,0 +1,60 @@
+@use "../../../sass/mixins";
+@use "../../../sass/settings";
+@use "../../../sass/typography";
+
+.form-item {
+  // @include blg-spacing("bottom", "small", "margin");
```
discodavey, line 6: We are not using blg-spacing on epp. We need to add margin 24px to the bottom of this.

src/components/atoms/forms/form-item.scss
```diff
@@ -0,0 +1,60 @@
...
+
+  & > :last-child {
+    @include mixins.margin(0, "bottom"); // Margins don't collapse on <select>s
+  }
+}
+
+.form-item__label_container {
+  @supports (display: flex) {
+    display: flex;
```
discodavey, line 15: Do we need to use flex on the default container?

src/components/atoms/forms/form-item.scss
```diff
@@ -0,0 +1,60 @@
...
+    justify-content: space-between;
+  }
+}
+
+.form-item__label {
+  @include mixins.margin(4, "bottom");
+
+  // @include typeg-form-label();
```
discodavey, line 23: we can use the font-size-and-vertical-height mixin and just add the bold etc to this

src/components/atoms/forms/text-field.scss
```diff
@@ -0,0 +1,31 @@
...
+  border: 1px solid settings.$divider;
+  border-radius: 3px;
+  color: settings.$color-text;
+  display: block;
+  font-family: settings.$font-family-secondary;
+  width: 100%;
+
+  &:focus {
+    border: 1px solid settings.$color-text-placeholder;
+    outline: 0;
```
discodavey, line 25: removing this will remove the browsers default focus style. Also in Safari it will not follow the shape of the border-radius

src/components/atoms/forms/form-item.scss
```diff
@@ -0,0 +1,60 @@
...
+  @include mixins.padding(24, "top");
+  @include typography.font-size-and-vertical-height(12, 24);
+
+  color: #757575;
+  display: inline-block;
+  font-family: settings.$font-family-secondary;
+}
+
+.form-item__message {
+  // @include typeg-form-message();
```
discodavey, line 48: we can use the font-size-and-vertical-height mixin and add the extra bits

PR #307, opened 2023-01-05
Share modal content

discodavey:


src/components/atoms/clipboard/clipboard.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const [copied, setCopied] = useState<boolean>(false);
+
+  useEffect(() => setShowButton(supportsClipboardAPI()), []);
+
+  return (
+    <div className="clipboard-container">
+      <div className="form-item">
+          <input type="input" className="text-field" value={text} />
+
+          {(showButton ? <button className={`button button--default button-clipboard${copied ? ' copied' : ''}`}
```
scottaubrey, line 22: Should we not be working to make the <Button> atom be applicable in all places, rather than reusing the <button> JSX element and importing styles everywhere?

scottaubrey, line 22: Should we not be working to make the <Button> atom be applicable in all places, rather than reusing the <button> JSX element and importing styles everywhere there is a button?

src/components/atoms/modal/modal.tsx
```diff
@@ -21,10 +23,12 @@ export const Modal = ({ modalTitle, modalContent }: Props): JSX.Element => {
     <button className="modal-button" onClick={() => setShowModal(!showModal)}>Modal Link</button>
```
scottaubrey, line 23: Same comments as before about buttons, but for Modal, I'm assuming we don't want to be including a button in the component, rather having the ability to open and close modals based on an external event of any description? Again, is this planned for later still?

src/components/atoms/socials/socials.stories.tsx
```diff
@@ -0,0 +1,18 @@
+import { ComponentStory, ComponentMeta } from '@storybook/react';
+import { Socials } from './socials';
+
+export default {
+  title: 'Atoms/ Socials',
```
scottaubrey, line 5: nitpick: probably don't need the space after the forward slash

src/components/atoms/socials/socials.tsx
```diff
@@ -0,0 +1,43 @@
...
+  redditUrl: string,
+};
+
+export const Socials = ({
+  emailUrl, twitterUrl, facebookUrl, linkedinUrl, redditUrl,
+}: Props): JSX.Element => (
+  <div className="socials-container">
+    <ul className="socials-sharers">
+      <li>
+        <a className="socials-sharer email" href={ emailUrl } target="_blank" rel="noopener noreferrer" aria-label="Share by Email">
```
scottaubrey, line 17: nitpick: spacing around these `{}` braces, but not the others below

src/components/atoms/socials/socials.tsx
```diff
@@ -0,0 +1,43 @@
...
+  linkedinUrl: string,
+  redditUrl: string,
+};
+
+export const Socials = ({
+  emailUrl, twitterUrl, facebookUrl, linkedinUrl, redditUrl,
+}: Props): JSX.Element => (
+  <div className="socials-container">
+    <ul className="socials-sharers">
+      <li>
```
scottaubrey, line 16: Would it be better to make these optionals, then only display the ones we have passed a URL for?

scottaubrey (changes requested):
From a design and styling perspective I've commented on some small minor things, a few nitpicks that are optional, but overall it does what it does fine to me.

Most of the remaining comments are around functionality and how we are going to use this - at the moment if we dropped this into a page it would be unusable due to the button styling, and we can't wire it up to another button or click anywhere else yet. I'm torn on whether it's better to tackle that sooner so that different modal content can benefit from it just working, or after you put together all the separate bits of modal content. It just whether that complexity is worth tackling soon, or just swallow the rework later. I'm happy to help work through pros and cons of each, but as long as we're aware it still needs doing I don't mind which.

discodavey:
@scottaubrey Are we happy with this now or does anything else need to be added for the share modal?

nlisgo:
> @scottaubrey Are we happy with this now or does anything else need to be added for the share modal?

Are you happy with it? Are you requesting a review now?

discodavey:
> > @scottaubrey Are we happy with this now or does anything else need to be added for the share modal?
> 
> Are you happy with it? Are you requesting a review now?

@nlisgo I spoke to @scottaubrey about this. I was meaning from functionality point of view.

When we bring in the form fields then we can refactor the clipboard text field.

discodavey:
This is now down for review.

src/components/atoms/clipboard/clipboard.scss
```diff
@@ -0,0 +1,51 @@
...
+    grid-row: 1;
+  }
+
+  &:focus {
+    border: 1px solid black;
+    outline: 0;
+  }
+}
+
+.text-field--textarea {
```
nlisgo, line 49: We have not introduce anything with this class so should remove.

src/components/atoms/clipboard/clipboard.scss
```diff
@@ -0,0 +1,51 @@
+@use "../../../sass/settings";
+@use "../../../sass/typography";
+@use "../../../sass/mixins";
+
+.clipboard-container {
+  width: 100%;
+}
+
+.form-item {
```
nlisgo, line 9: This probably belongs in it's own component.

discodavey:
@scottaubrey Can you please go through this so we can get this done.

scottaubrey (approved):

PR #310, opened 2023-01-06
Refactor modal to improve styles and remove trigger from template

nlisgo:
This progresses the modal component in the following ways:

- improves the styling to ensure that the close button is inline with the title matching designs.
- ensures that there is no trigger for opening the modal within the modal component but provides an example (with tests) on how the trigger behaviour can be assigned to an element in the DOM.
- will simplify the [PR for the share modal](https://github.com/elifesciences/enhanced-preprints-client/pull/310)

### DIFF
--- a/src/components/atoms/modal/modal.scss
+++ b/src/components/atoms/modal/modal.scss
@@ -13,6 +13,9 @@
 }
 
 .modal-content {
+  @include mixins.margin(0);
+  @include mixins.padding(36);
+
   background-color: #fff;
   border-radius: 3px;
   box-shadow: 0 0 20px rgb(0 0 0 / 22%);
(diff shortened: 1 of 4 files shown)


src/components/atoms/modal/modal.scss
```diff
@@ -22,29 +25,48 @@
...
 
   @media only all and (min-width: settings.$breakpoint-small) {
-    margin: 24px;
+    @include mixins.margin(24);
   }
 
   @media only all and (min-width: settings.$breakpoint-medium) {
-    margin: 180px auto 24px;
+    @include mixins.margin(180, "top");
+    @include mixins.margin(auto, "right");
```
discodavey, line 33: Auto is for left and right

src/components/atoms/modal/modal.scss
```diff
@@ -22,29 +25,48 @@
...
+    @include mixins.margin(24, "top");
```
discodavey, line 34: You have two margin-tops

scottaubrey (approved):

src/components/atoms/modal/modal.tsx
```diff
@@ -1,30 +1,35 @@
...
-    <div onClick={(event) => clickHandler(event)} className={`modal-container${showModal ? ' modal-content__show' : ''} `}>
+    <div onClick={(event) => { if (onModalClose !== undefined && clickDetectedOutsideOfModal(event)) { onModalClose(); } }} className={`modal-container${open ? ' modal-content__show' : ''} `}>
       <div ref={contentRef} className="modal-content">
-        <button className="modal-content__close-button" onClick={closeModal}>Close</button>
         <div className="modal-content__block">
-          <h6>{modalTitle}</h6>
-          {modalContent}
+          <div className="modal-content__top">
+            <h6 className="modal-content__title">{modalTitle}</h6>
+            {onModalClose !== undefined ? (<button className="modal-content__close-button" onClick={onModalClose}>Close</button>) : ''}
```
discodavey, line 30: Are we going to use the button component for this? Just wondering if not as it's different from the other style of buttons as the icon is on the left where as this is on the right

discodavey (approved):

discodavey:
Merging as agreed with @nlisgo

PR #313, opened 2023-01-09
Button refactor

discodavey:


nlisgo:
@scottaubrey could you take a look at why the preview environment is not working?

scottaubrey:
> @scottaubrey could you take a look at why the preview environment is not working?

It is working now

scottaubrey:
> > @scottaubrey could you take a look at why the preview environment is not working?
> 
> It is working now

I didn't do anything BTW - it was just working

discodavey:
@chris-huggins can you have a look at the buttons i.e. download button, sign up and on the page not found to make sure they all look good as we have done some changes with the coding but they should be as before

nlisgo:
@scottaubrey thanks for checking anyway. Can you remind me how to access storybook on the preview instance please?

scottaubrey:
> @scottaubrey thanks for checking anyway. Can you remind me how to access storybook on the preview instance please?

Sorry, not updated the links created here yet - it's the same as the preview link, but with `-storybook` on the end, e.g. https://pr-313--epp-storybook.elifesciences.org/

scottaubrey (dismissed):

chris-huggins:
@discodavey I'm gradually working on removing ALLCAPS from buttons, starting with the action buttons on EPP and journal articles (the light blue/grey background). But I'm seeing them with ALLCAPS in [storybook](https://pr-313--epp-storybook.elifesciences.org/).  Could you please make them Title Case? And we can eventually update the blue primary buttons used elsewhere in journal to title case, but let's not worry about it yet.
image

nlisgo (dismissed):

chris-huggins:
Sorry @nlisgo I'm getting 404 errors for those links. Can you steer me in the right direction? I don't mind just checking via storybook for now

scottaubrey (approved):
I'm OK merging this in, though I'm really not enamoured with the `classVariant` function. It doesn't read or help me understand it any better than the ternary, but that might just be me.

Looking through it, I thought of some other improvements, that are optional for merging, but you might want to do them:
- `classVariant` name - could something like `classNameVariant` be better? especially in the context (TS/JSX) where "class" means 2 different things.
- `button--icon` could be a mixin, used in each button icon class, stopping the issue of weird looking output if you forget `button--icon`
- Setting argTypes in storybook to reduce the number of stories/examples given, e.g. add this to the default export in `button.stories.tsx`:
  ```
    argTypes: {
    variant: {
      options: [
        '',
        'action',
      ],
      control: { type: 'radio', labels: { '': 'None', action: 'Action' } },
    },
    iconName: {
      options: [
        '',
        'download',
        'share',
        'citation',
        'follow',
      ],
      control: { type: 'select', labels: { '': 'None' } },
    },
  },
  ```
  Reload and have a play with the controls

nlisgo:
I'm picking up the feedback on default argTypes for button. That should be great. Also changing `classVariant` to `classNameVariant` we can change this again later if further use of it surfaces a better name.

@scottaubrey I'm unclear on the value of the mixin as the `button--icon` will be set when the `iconName` is supplied. You're not getting an icon without supplyin an iconName! A brief call on this should clear up any uncertainty.

src/components/atoms/button/button.tsx
```diff
@@ -1,13 +1,17 @@
 import './button.scss';
+import { classNameVariant } from '../../../utils/class-name-variant';
 
 type ButtonProps = {
-  iconName: string,
+  iconName?: 'download' | 'share' | 'citation' | 'follow',
+  variant?: 'action',
```
nlisgo, line 6: I feel like we could abstract these possible values in a way that can serve the storybook options. But will have a think about this one and perhaps take it on in 10% time.

scottaubrey: Agreed. They would feel better as an Enum perhaps - each new icon needs a new value.

scottaubrey:
> I'm picking up the feedback on default argTypes for button. That should be great. Also changing `classVariant` to `classNameVariant` we can change this again later if further use of it surfaces a better name.
> 
> @scottaubrey I'm unclear on the value of the mixin as the `button--icon` will be set when the `iconName` is supplied. You're not getting an icon without supplyin an iconName! A brief call on this should clear up any uncertainty.

That is the behaviour in code yes, but that's the justification for asking if we needed two separate classes given there is only even the outcome of having 1 or 3 classes, why not 1 or 2 classes. Then, the ClassVariant doesn't have a higher WTF  factor when the string is a space separated list of only one class getting variant-ed. Happy to chat if that doesn't make sense.

src/components/atoms/button/button.test.tsx
```diff
@@ -2,11 +2,29 @@ import { render, screen } from '@testing-library/react';
...
+    expect(screen.getByText('button text').getAttribute('href')).toStrictEqual('bbc.co.uk');
+  });
+
+  it('renders with the provided text, iconName and variant', () => {
+    render(<Button text='button text' iconName='follow' variant='action' url='bbc.co.uk' />);
 
     expect(screen.getByText('button text')).toBeInTheDocument();
-    expect(screen.getByText('button text')).toHaveClass('dark_mode');
+    expect(screen.getByText('button text')).toHaveClass('button--icon-follow');
+    expect(screen.getByText('button text')).toHaveClass('button--action');
```
nlisgo, line 27: I considered adding a test to check that only acceptable values were supplied to iconName and variant but the tests don't seem to fail if I supply string that isn't expected such as:

```
render(<Button text='button text' iconName='unknown' variant='unknown' url='bbc.co.uk' />);
```

nlisgo (approved):

nlisgo:
@scottaubrey I believe I've addressed all of your feedback. @discodavey can you check you are happy? Hoping we can merge this in today so @discodavey can continue work on: https://github.com/elifesciences/enhanced-preprints-client/pull/307

scottaubrey (approved):

scottaubrey (approved):

PR #317, opened 2023-01-10
Make familyNames and givenNames optional

discodavey:


scottaubrey:
This has worked! 

Observe: 
https://prod--epp.elifesciences.org/preview/10.1101/2022.11.10.516056
vs
https://pr-317--epp.elifesciences.org/preview/10.1101/2022.11.10.516056

scottaubrey (approved):
This works!

Ideally, we'd maybe go with blank string on givenName, but a fallback when given and family names are empty, something like "unknown" or "anonymous" or something.

However, I think this is no worse than other areas that would display blank strings when the both are empty, and this doesn't blow up with the content linked. I'm happy to go with this to fix the bug.

PR #318, opened 2023-01-11
Set z-index of .tabbed-navigation to ensure that the jump menu passes behind it

nlisgo:


nlisgo:
@scottaubrey :wave: sorry the preview environment is not working. It was. But isn't. Please help but show me how you resolve it.

scottaubrey:
> @scottaubrey 👋 sorry the preview environment is not working. It was. But isn't. Please help but show me how you resolve it.

We didn't get chance before I finished, but I'm happy to either tomorrow, or I can try lay out some steps here.

In this instance, I highly suspect the image tag being looked for doesn't exist, and I suspect it's because the PR is out of date. I've done the merge to see if it fixes it. I got this from the output from the (at time of writing) [latest preview env run](https://github.com/elifesciences/elife-flux-cluster/actions/runs/3903994019/jobs/6669149986): `skipping PR 318, image doesn't exist yet`

nlisgo:
Thanks a lot @scottaubrey. Preview instance is up again!

src/sass/_mixins.scss
```diff
@@ -107,3 +107,7 @@
   position: static;
   width: auto;
 }
+
+@mixin z-index($value) {
```
nlisgo, line 111: This mixin ensures that z-index is a multiple of 10.

scottaubrey: Nice solution to the enforced multiples of 10.

scottaubrey (approved):

src/sass/_mixins.scss
discodavey: I think this is over-engineering as the amount of times we use z-index is only going to be a small amount.

nlisgo: Maybe I can adapt the mixin to throw a warning if we use it too much. :). I will let it in for now and we can remove/refactor if it proves to be annoying in practice.

Thanks for the review.

discodavey: It's not annoying there is just no need for it. I put z-index in the same pot as position absolute where you only use it when you only need to. Very rarely.

PR #321, opened 2023-01-13
Start building images tagged with pr_id, and only post Preview URL if  detected 200 status code

scottaubrey:
We recently switch the env build script to post the latest digest instead of relying on the commit hash, and to check for pr_id tag name This switches the image tag to be pr_id, which:
- reduces the number of tags in the CR view
- allows for there to not be a flip/flop when checking for the latest image (we just update the digest instead)
- pr_id is slightly more consistent that the sha, as that changes with pushes to master (without triggering the build).

Finally, we stall the pipeline for 30 mins waiting for the Environment to appear before we publish the Preview Environment. This should also make the developer experience slightly better, rather than having to infer and wait after the env URL is posted.

### DIFF
--- a/.github/workflows/build-preview.yaml
+++ b/.github/workflows/build-preview.yaml
@@ -32,13 +32,13 @@ jobs:
           target: prod
           load: true
           tags: |
-            ${{ env.CLIENT_IMAGE_REPO }}:preview-${{ github.ref_name }}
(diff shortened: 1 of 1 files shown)


nlisgo (approved):
LGTM: merge when green!

nlisgo:
I can't understand why your changes are causing the failure in CI.

scottaubrey:
The failure look unrelated - network level failure between github and the source server in azure maybe?

PR #323, opened 2023-01-13
Introduce browser tests for institutions

nlisgo:


browser-tests/content-header.spec.ts
```diff
@@ -20,8 +20,8 @@ test.describe('content header', () => {
   });
 
   test('has correct authors', async () => {
-    await contentHeader.assertAuthorExists('Brain');
-    await contentHeader.assertAuthorExists('Pinky Mouse');
+    await contentHeader.assertAuthorsExists('Brain');
```
nlisgo, line 23: these typos are unrelated but was hoping that they didn't warrant a separate PR. I can split them out if valuable.

will-byrne (approved):

PR #326, opened 2023-01-16
Reference build

discodavey:


src/components/atoms/reference-list/reference-list.tsx
```diff
@@ -1,46 +1,13 @@
...
-  );
-};
-
-export const ReferenceList = ({ references }: { references: Reference[] }): JSX.Element => (
+export const ReferenceList = ({ references }: { references: ReferenceData[] }): JSX.Element => (
   <section>
     <Heading id="references" headingLevel={1} content="References" />
     <ul className="reference-list">
-      {references.map((reference, index) => <ReferenceListItem key={index} reference={reference} />)}
+      {references.map((reference, index) => <li key={index}><Reference reference={reference} doiIsLink={true} /></li>)}
```
nlisgo, line 10: Is this the right time to move reference-list to molecules as it is dependant on an atom?

src/components/atoms/reference/reference.tsx
```diff
@@ -0,0 +1,37 @@
+import { Author, Reference as ReferenceData } from '../../../types';
+import './reference.scss';
```
nlisgo, line 2: Let's add a test for this. As this is a refactor it should be easy to use the test for reference-list as a starting point.

src/components/atoms/reference-list/reference-list.tsx
discodavey: Yes, because we need it in the cite modal

src/components/atoms/reference/reference.tsx
discodavey: This is not ready for review as yet

nlisgo (changes requested):
Looks good. Just needs a couple of tweaks.

src/components/atoms/reference/reference.tsx
nlisgo: ok, that's fine. stepping away.

discodavey:
> Looks good. Just needs a couple of tweaks.
We are still working on this so is not ready for review.

will-byrne (approved):

PR #361, opened 2023-02-03
Introduce workflow to publish manuscript

nlisgo:
If you use workflow from master then the changes to manuscripts.json will be made and introduced as a PR. 

If however you use a workflow from a feature branch the changes will be committed to that branch. This allows the GitHub action to be used to collect all new manuscripts into a single PR reducing the burdens on deployment. 

If you are happy with the approach the code is safe as it doesn’t impact the application. PR’s should still be manually reviewed for now and tests can be added to improve confidence.


scottaubrey (approved):
I give it a tentative approval.

The script works, and seems to work quite well at that. 

The extra dependancies make it a little less easy to remove when the time comes, but not substantially so. Same for the package-config ts-node config.

I haven't tested the GH workflow but it looks OK to me.

As we've said a few times, it's not automatically doing anything, so as long as we always eyeball the result for the time being, i'm OK with proceeding.

scottaubrey (approved):

PR #363, opened 2023-02-03
refactor grid so action buttons collapse correctly

discodavey:


chris-huggins:
@discodavey I think we should still try to remove the right margin/padding on the alternate buttons (i.e. Cite) so that we can make the aside column narrower still (down to 133px). That would mean it matches the designs and then the text at the top of the aside will fit a bit better. Happy to pair on it if needed.

discodavey:
@chris-huggins I feel by removing margins on the alternate buttons that if we introduce another button that was slightly shorter or longer it could cause the buttons to butt up to each other without the margin.

All depends if we are looking at future proofing this. If like other components that we have built where future proofing is not being bothered with then I am happy to do this,

discodavey:
@chris-huggins I have reduced the max-width and alternated the margin-right on the action buttons. Will let you know when it's ready to review

chris-huggins:
@discodavey Thanks. I was going to say we should be fine with buttons. I only anticipate us adding a fourth after the "Share" button for "Follow". Any other changes will need some consideration so we can address any problems caused by margins then. I think it's more important to ensure the text fits correctly than worrying about changes to buttons.

scottaubrey (approved):

PR #366, opened 2023-02-03
Order timeline events chronologically (descending)

fred-atherden:
fixes elifesciences/enhanced-preprints-issues#458


nlisgo:
@FAtherden-eLife you legend! I will take a look at this when I get some time.

src/components/molecules/timeline/timeline.test.tsx
```diff
@@ -11,9 +11,12 @@ describe('Timeline', () => {
       { name: 'event2', date: '2002-02-14' },
     ]}/>);
 
+    const firstEvent = screen.getByText('14 February 2002');
+    const secondEvent = screen.getByText('13 January 2001');
+    expect(firstEvent).toBeInTheDocument();
+    expect(secondEvent).toBeInTheDocument();
+    expect(firstEvent.compareDocumentPosition(secondEvent)).toBe(4);
```
nlisgo, line 18: I will test this out locally before committing but if this test would fail before your fix in `src/components/molecules/timeline/timeline.tsx` then it's always a good idea to:

- commit the test change first
- create PR
- CI should fail
- commit the fix
- CI should pass

src/components/molecules/timeline/timeline.tsx
```diff
@@ -20,7 +20,7 @@ export const Timeline = ({ events }: TimelineProps): JSX.Element => (
   <div className="review-timeline">
     <dl className="review-timeline__list">
       {
-        events.map((entry, index) => (
+        events.sort((a, b) => new Date(b.date).getTime() - new Date(a.date).getTime()).map((entry, index) => (
```
nlisgo, line 23: Switched to this which has the same expected effect but this approach is slightly more robust.

nlisgo (approved):

nlisgo:
Thanks for the PR @FAtherden-eLife. Will merge this in on Thursday morning.

FAtherden-eLife:
Good to know - thanks!

nlisgo:
Merging when green after pulling in changes from upstream

PR #370, opened 2023-02-07
fixes very large inline image equations in 10.1101/2022.12.15.520660

FAtherden-eLife:
[10.1101/2022.12.15.520660](https://prod--epp.elifesciences.org/preview/10.1101/2022.12.15.520660#fig1):
Screenshot 2023-02-07 at 08 26 40
<br />
<br />
**changed to:**
Screenshot 2023-02-07 at 08 27 16


src/components/atoms/figure/figure.scss
```diff
@@ -5,7 +5,7 @@
 .figure {
   margin: 0;
 
-  img {
+  >img {
```
discodavey, line 8: spacing needed between greater than and img

scottaubrey (approved):
LGTM, thanks @FAtherden-eLife !

This does fix it for that paper very nicely, and is a neat way to avoid treating all imgs inside a figure styled as the main event. 

@discodavey advised that in general we try to avoid having these kind of specific selectors, but we couldn't think of a way without writing all the styles again for figurecaption, i.e.:

```
.figure img { max-width: 100% /* etc */ } 
.figurecaption img { max-width: initial /* undo */ }
```

Which felt like we'd be maintaining two lots of styles for no real benefit.

discodavey:
I was happy with what was there. Just needed spacing ``` > img```

scottaubrey (dismissed):
LGTM @FAtherden-eLife , thanks very much!

@discodavey advised that we try to avoid specific selectors like this, but the only other way we could come up with requires twice as much SASS code, by targetting the `.figurecaption` to undo the styles in `.figure`, i.e.:

```
.figure img { max-width:100% /* etc */ }
.figurecaption img { max-width: initial /* etc undoing the same above */ }
```

This change, while unideal, is neat, and in theory we're not expecting many other uses for img tags under `<figure>`. So I think we're good to merge. Feel free when you're ready

scottaubrey:
Apologies for the noise - Github's having a moment this morning. It didn't show the first approval the first time.

scottaubrey (approved):

FAtherden-eLife:
Good to know! Thanks for the review, and no problem about noise.

PR #384, opened 2023-02-13
Misc tweaks and fixes for docker-compose api-fetch-data service

scottaubrey:
add longer timeout, fix other timeout values, change branch and make sure nginx doesn't start until the service has imported all manuscripts

@discodavey This is the branch you might want to try locally to see if your epp-client comes up without 500 errors


discodavey (approved):
Both localhost 6006 and 8080 now working correctly locally.

PR #407, opened 2023-02-23
remove alt text added for testing in early phase epp

will-byrne:


scottaubrey (approved):

discodavey:
Nice bit of codey @will-byrne

PR #422, opened 2023-03-03
update terms bolded in eLife assessment

fred-atherden:


nlisgo:
We need to create a ticket for this. There are 2 issues here:

- addition of the new term 'exceptional'. If all this was doing was that we could probably merge it in.
- allow search for term to be case insensitive, with accompanying test.


cc @QueenKraken

src/components/atoms/review-content/review-content.tsx
```diff
@@ -1,10 +1,14 @@
 import Link from 'next/link';
 import './review-content.scss';
 
+/* eslint-disable */
```
will-byrne, line 4: Do we need to disable the linter here?

fred-atherden: Adding new lines throws `no-trailing-spaces`.
Not having new lines throws `max-len` (333 > 240)

Looks like @nlisgo is suggesting a different approach above, so have tentatively proposed a different one below.

E-Kemp:
Thanks for this Fred, we've gone ahead and made some changes to reflect the need to be case insensitive with the regex (and updated the terms): https://github.com/elifesciences/enhanced-preprints-client/pull/443

PR #431, opened 2023-03-06
Add min-width to fix DNA pushing grid out over side-section

discodavey:


scottaubrey (dismissed):
Seems to work well for 84792

discodavey:
This needs to be tested locally before pushing to production. You will need to copy and paste DNA in to the inspect to make it work as there is no preview available.

This is re Freds comment in https://github.com/elifesciences/issues/issues/8125

scottaubrey:
I've checked out the storybook content too. That's fine.

@discodavey - I've checked the actual content posted for article msid 84792 which has the DNA string in it. Switching between branches between this and master redraws the content out of bound then in bound - presumably that is sufficient, and a way better test than adding it via the dev console?

scottaubrey (approved):

discodavey:
> I've checked out the storybook content too. That's fine.
> 
> @discodavey - I've checked the actual content posted for article msid 84792 which has the DNA string in it. Switching between branches between this and master redraws the content out of bound then in bound - presumably that is sufficient, and a way better test than adding it via the dev console inspect?

The user won't be switching between the branch and master as they will be only seeing one version.

Shall we get anyone else to test this as we do not have preview so Ash and Chris cannot test?

scottaubrey:
> > I've checked out the storybook content too. That's fine.
> > @discodavey - I've checked the actual content posted for article msid 84792 which has the DNA string in it. Switching between branches between this and master redraws the content out of bound then in bound - presumably that is sufficient, and a way better test than adding it via the dev console inspect?
> 
> The user won't be switching between the branch and master as they will be only seeing one version.

Sorry Dave, I meant I was testing by switch branch.

`git checkout master`:
Screenshot 2023-03-06 at 3 05 47 pm

`git checkout main-content-area-width-issue`:
Screenshot 2023-03-06 at 3 06 25 pm


> Shall we get anyone else to test this as we do not have preview so Ash and Chris cannot test?

You can do, but we'll need to get you up and running locally, using the steps I gave you this morning.

PR #440, opened 2023-03-07
refactor tab tests

will-byrne:


nlisgo (approved):
This is a better foundation. The todo's can be picked up in a followup ticket.

https://github.com/elifesciences/enhanced-preprints-issues/issues/516

PR #455, opened 2023-03-15
add pdf urls for various RPs

fred-atherden:
Adding via the workflow would cause numerous deployments, so adding here all in one PR.


nlisgo (approved):
Happy for this to be merged

PR #481, opened 2023-03-20
Slim down the EPP-client container

scottaubrey:
The EPP client container is one of the biggest on the cluster, taking up 3.2GB. To make matters worse - we deploy most of that again for the storybook container.

We can reduce the size considerably by simply ignore newly added data files, and building node_modules in a separate build step and only copying the output


E-Kemp (approved):
It works :)

PR #497, opened 2023-03-23
Crossref bibtex

E-Kemp:
Closes https://github.com/elifesciences/enhanced-preprints-issues/issues/410


next.config.js
```diff
@@ -18,6 +18,10 @@ const nextConfig = {
         source: '/reviewed-preprints/:path((?!\\d+v{0,1}\\d*))',
         destination: '/:path*',
       },
+      {
+        source: '/reviewed-preprints/:msid(\\d+v{0,1}\\d*).:format(bib|ris)',
```
nlisgo, line 22: Should we remove the `ris` until we have implemented that on server?

src/components/molecules/article-status/article-status.tsx
```diff
@@ -49,13 +50,22 @@ export const ArticleStatus = ({
       <Modal modalTitle={'Share this article'} open={showShareModal} onModalClose={() => setShowShareModal(false)} modalLayout="share">
         <div className="form-item">
           <input readOnly={true} type="input" className="text-field text-field--clipboard" value={`https://doi.org/${doi}`} />
-          <Clipboard text={`https://doi.org/${doi}`} />
+          <div className="clipboard-container">
```
nlisgo, line 53: Why has this div been moved outside of the clipboard.tsx?

src/pages/api/citations/[...filename].page.ts
```diff
@@ -0,0 +1,32 @@
...
+export default async (req: NextApiRequest, res: NextApiResponse) => {
+  const { filename } = req.query;
+  if (filename !== undefined) {
+    const { name, ext } = path.parse(Array.isArray(filename) ? filename[0] : filename);
+
+    const manuscript = getManuscript(config.manuscriptConfigFile, name);
+
+    if (manuscript !== undefined && name !== undefined && (ext === '.bib' || ext === '.ris')) {
+      const extReq = await axios.get(
+        `${config.apiServer}/api/citations/${manuscript.preprintDoi}/${ext === '.bib' ? 'bibtex' : 'ris'}`,
```
Comment (scottaubrey) [inline src/pages/api/citations/[...filename].page.ts:16]: I'm pretty certain this needs to be a constructed DOI for the RPP, not the preprint's DOI

scottaubrey (changes requested):

src/components/molecules/article-status/article-status.tsx
E-Kemp: Because it was affecting the positioning of the button within the modal, which made it a pain for us to use it in a group of buttons

next.config.js
E-Kemp: I can remove it and add a reminder of some sort for when we add .ris support?

src/components/atoms/button/button.tsx
```diff
@@ -2,20 +2,21 @@ import './button.scss';
...
 export const buttonIconNames: Array<string> = ['download', 'share', 'citation', 'follow'];
-export const buttonVariants: Array<string> = ['action', 'clipboard'];
+export const buttonVariants: Array<string> = ['action', 'clipboard', 'cite-download'];
 
 type ButtonProps = {
   iconName?: typeof buttonIconNames[number],
   variant?: typeof buttonVariants[number],
   text: string,
   url?: string,
+  download?: boolean,
```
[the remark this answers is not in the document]
E-Kemp: Good spot, the button now has the download attribute enabled (telling the browser the link is a download rather than a navigation location)

src/pages/api/citations/[...filename].page.ts
Comment (E-Kemp) [reply src/pages/api/citations/[...filename].page.ts]: Good spot, I've lifted the logic that gets the eLife DOI when we fetch data from server and re-used it

Comment (scottaubrey) [reply src/pages/api/citations/[...filename].page.ts]: Yup tested on the latest preview env, works like a charm!

src/pages/api/citations/[msid]/bibtex.page.ts
```diff
@@ -0,0 +1,35 @@
+import axios from 'axios';
```
Comment (scottaubrey) [inline src/pages/api/citations/[msid]/bibtex.page.ts:1]: Could we use fetch? We deliberately tried to use fetch in other areas to keep client bundle size down, but that doesn't matter here. Still, it feels like it would be more consistent to keep using fetch here instead of importing axios

src/pages/api/citations/[msid]/bibtex.page.ts
```diff
@@ -0,0 +1,35 @@
...
+import type { NextApiRequest, NextApiResponse } from 'next';
+import { getManuscript, getRppDoi } from '../../../../manuscripts';
+import { config } from '../../../../config';
+
+export default async (req: NextApiRequest, res: NextApiResponse) => {
```
Comment (scottaubrey) [inline src/pages/api/citations/[msid]/bibtex.page.ts:6]: Can this be restructured to return errors early so the happy path is the final return, rather than multi-level conditional nesting?

scottaubrey (approved):
I've posted a few comments, nothing massive just a bit of restructuring

scottaubrey (approved):

nlisgo:
Hold off on merge. Just getting an error on the bibtex download locally and I want to ensure it's just a local issue.

PR #511, opened 2023-03-25
Add the storybook a11y addon

nlisgo:


scottaubrey (dismissed):

discodavey:
They have this in the inspect on Chrome and Firefox already

nlisgo:
> They have this in the inspect on Chrome and Firefox already

Makes it harder to ignore and also allows us to integrate accessibility checks with CI. Think this is a great little add-on. Helps us keep accessibility as a priority.

nlisgo:
I've reset to the approved commit and preserved the additional work to demonstrate how it could impact ci, if we wish.

PR #517, opened 2023-03-27
Clean up dependencies and devDependencies

nlisgo:


package.json
```diff
@@ -3,20 +3,12 @@
...
-    "@types/node": "^18.0.0",
-    "@types/react": "^18.0.0",
-    "@types/react-dom": "^18.0.0",
     "crc-32": "^1.2.2",
     "next": "^13.0.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-scripts": "5.0.1",
     "sass": "^1.54.5",
-    "typescript": "^4.4.2",
```
scottaubrey, line 19: Production image run ts-node, which I assume requires typescript? We don't ahead of time compile to js anyway.

will-byrne (dismissed):

PR #520, opened 2023-03-28
Added no-only-tests eslint plugin

E-Kemp:
Resolves https://github.com/elifesciences/enhanced-preprints-issues/issues/579


package.json
```diff
@@ -11,6 +11,7 @@
     "@types/react": "^18.0.0",
     "@types/react-dom": "^18.0.0",
     "crc-32": "^1.2.2",
+    "eslint-plugin-no-only-tests": "^3.1.0",
```
will-byrne, line 14: this should be a `devDependency` should it not?

will-byrne (approved):

PR #533, opened 2023-03-31
Mastadon share button

E-Kemp:
Closes https://github.com/elifesciences/enhanced-preprints-issues/issues/585


discodavey:
Should these icons be on the same line?

Screenshot 2023-03-31 at 12 33 35

src/components/atoms/socials/socials.scss
```diff
@@ -4,7 +4,6 @@
 
 .socials-container {
```
discodavey, line 5: If we are not using this class now then we need to remove it

nlisgo:
@chris-huggins are you able to approve?

https://pr-533--epp.elifesciences.org/reviewed-preprints/80494

chris-huggins:
@nlisgo Looks good to me, thanks.

nlisgo:
> Should these icons be on the same line? This is at 320 breakpoint
> 
> Screenshot 2023-03-31 at 12 33 35

@chris-huggins please raise this as a ticket if not working as expected with guidance on how it should be working. Thanks

chris-huggins:
@nlisgo Yes Dave is right. I've raised a ticket here:  https://github.com/elifesciences/enhanced-preprints-issues/issues/586

PR #546, opened 2023-04-04
fix alignment for social sharers in modal

discodavey:


scottaubrey (approved):
Seems to have fixed the socials on the same line at 320 width

PR #548, opened 2023-04-04
[SNAG] Use preprint DOI for citation call if eLife DOI doesn't exist

E-Kemp:


E-Kemp:
> Need a couple more replacements if published: Before this PR:
> 
> https://elifesciences.org/reviewed-preprints/84141.bib
> 
> ```
> @article{Zhang_2023,
> 	doi = {10.7554/elife.84141.1},
> 	url = {https://doi.org/10.7554/elife.84141.1},
> 	year = 2023,
> 	month = {mar},
> 	publisher = {{eLife} Sciences Publications, Ltd},
> 	author = {Tony Zhang and Matthew Rosenberg and Pietro Perona and Markus Meister},
> 	title = {Endotaxis: A neuromorphic algorithm for mapping, goal-learning, navigation, and patrolling}
> }
> ```
> 
> After:
> 
> https://pr-548--epp.elifesciences.org/reviewed-preprints/84141.bib
> 
> ```
> @article{Zhang_2021,
> 	doi = {10.7554/eLife.84141.1},
> 	url = {https://doi.org/10.1101/2021.09.24.461751},
> 	year = 2021,
> 	month = {sep},
> 	publisher = {Cold Spring Harbor Laboratory},
> 	author = {Tony Zhang and Matthew Rosenberg and Pietro Perona and Markus Meister},
> 	title = {Endotaxis: A neuromorphic algorithm for mapping, goal-learning, navigation, and patrolling}
> }
> ```
> 
> We need to change the citeKey (year), url, year, month and publisher. I completely overlooked this when I called this a golden snag but still believe this has legs until we can bring this into server.

This feels like we'll be closer to generating our own than generating one from crossref...

nlisgo:
> This feels like we'll be closer to generating our own than generating one from crossref...

Most of the complexity is in generating the title and authors so this approach still has legs but this definitely doesn't feel snaggy any more. Let's park.

nlisgo:
Alternatives have been suggested which might contribute to an acceptable solution. We could only present the bibtex citation download if the the service is available. Or we could just agree to not display the link a set number of days after publication.

nlisgo:
blocking until an acceptable approach is agreed. Was good to address some unknowns on this.

PR #550, opened 2023-04-04
Added RIS download button and routing

E-Kemp:
Closes https://github.com/elifesciences/enhanced-preprints-issues/issues/412


nlisgo (approved):
Only suggested improvement is to merge the `bibtex.page.ts` and `ris.page.ts` as they share so much code. Not blocking but could be done at some stage.

LGTM

chris-huggins:
@E-Kemp Can you help me with where I should be looking? I can't check locally and not seeing it on staging. Happy to just see it on your local version on a call if easier. Thanks.

discodavey:
The buttons do not collapse as would be expected. They need to go 100% at one of the breakpoints but would discuss with Chris if that is 480px or 768px

chris-huggins:
@E-Kemp Dave's right, ideally on smaller screens we would move to 100% width buttons stacked vertically instead of one or two wrapping below the copy to clipboard button.

We could replicate what we have in journal (screenshot 1 below) by doing this at a standard breakpoint, which would be 768px on EPP I think. If we do that I'd like to see the vertical space between buttons smaller than we have on journal currently (12px would be good)

However, it would be better if we could do something smarter where we only go 100% when the buttons run out of space on a single line. I don't know if something like container queries could be possible? Maybe you know a different way to achieve this?

In your PR that would be at 643px but it could be even narrower if we lose the right margin on the last button. (see screenshot 2 below)

Finally, I'd also love to see these EPP button styles (also where they're used in other places on EPP) tweaked very slightly to align the label better. Currently top and bottom padding are 13px (screenshot 3). I'd like to change them to 12px and 14px (screenshot 4) to bump the label by 1px. Such a small thing seems unnecessary to raise a ticket but let me know if it's needed. Thanks.

1. 
image


2. 

image

3. 
image

4.

image

chris-huggins:
@E-Kemp Here's a mockup/comparison at 600px wide to clarify what I'm hoping to avoid and what I'm hoping we can do. We could do this from the 768px breakpoint but if there's a way to do it just before the RIS button wraps that would be better. Thanks.

image

E-Kemp:
@chris-huggins I've made all the changes:

- In regards to the label padding I made the change to the _cite_ buttons, when I shifted the labels in the smaller buttons it looked a bit odd as that misaligned the text with in-line icons.
- The breakpoint is set manually rather than using one of our predefined breakpoints - I think it behaves really nicely!

nlisgo (approved):
Thanks for the input everyone. I think this meets the brief now.

Works well at the different breakpoints.

PR #554, opened 2023-04-04
add more context, description and the storybook preview URL to preview environments

scottaubrey:
fixes https://github.com/elifesciences/enhanced-preprints-issues/issues/417


E-Kemp (approved):
This works just like the EPP Preview link, great! ✨ 

Worth noting Storybook has an ongoing issue (see https://github.com/elifesciences/enhanced-preprints-issues/issues/573) but that doesn't concern this PR

PR #556, opened 2023-04-04
Potential fix for Module not found: Can't resolve 'fs' in '/opt/epp-client/src'

nlisgo:
Solution taken from: https://stackoverflow.com/questions/71158775/storybook-couldnt-resolve-fs

Not sure if it is all needed but working for me locally.


nlisgo:
I'm not going to pretend to know exactly why this is working but this is for me locally and in preview. My strong recommendation is that we merge this in and if we never see this issue again then accept the win. Open to feedback,

E-Kemp (approved):
Seems to be a common occurrence with Webpack, looks good to me!

PR #560, opened 2023-04-05
Add claim rendering

discodavey:


discodavey:
@scottaubrey Can we pair on this tomorrow morning?

src/utils/content-to-jsx.tsx
```diff
@@ -52,6 +52,12 @@ export const contentToJsx = (content: Content, index?: number): JSXContent => {
       return <li key={index}>{contentToJsx(content.content)}</li>;
     case 'List':
       return content.order === 'Ascending' ? <ol key={index}>{contentToJsx(content.items)}</ol> : <ul key={index}>{contentToJsx(content.items)}</ul>;
+    case 'Claim':
+      return <section key={index}>
+        {content.title ? <h3>{contentToJsx(content.title)}</h3> : ''}
+        <h4>{content.label}</h4>
+        {contentToJsx(content.content)}
```
fred-atherden, line 59: Suggest that either both label and title should be in the same header (this would be my preferred rendering), or that label should appear before title.

src/types/content.ts
```diff
@@ -74,6 +74,13 @@ type ListContent = {
   items: Array<ListItemContent>,
 };
 
+type ClaimContent = DecoratedContent & {
+  type: 'Claim',
+  claimType: 'Statement' | 'Theorem' | 'Lemma' | 'Proof' | 'Postulate' | 'Hypothesis' | 'Proposition' | 'Corollary',
+  label: string,
```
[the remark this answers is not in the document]
scottaubrey: > In jats the label element [can contain multiple elements](https://jats.nlm.nih.gov/archiving/tag-library/1.3/element/label.html) (not just a string)

That sounds reasonable.

> if a title has been provided but not a label in the source XML

@fred-atherden Does this also mean label should be an optional as well?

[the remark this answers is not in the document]
fred-atherden: Suggest that we make `label` optional too. It's optional in jats, and if no label or title is provided, then encoda produces json without it:

<details>
   <summary>Example</summary>

```json
{
      "type": "Claim",
      "id": "alg1",
      "content": [
        {
          "type": "Paragraph",
          "content": []
        },
        {
          "type": "Figure",
          "id": "alg1a",
          "content": [
            {
              "type": "ImageObject",
              "contentUrl": "461751v2_alg1.tif",
              "meta": {
                "inline": false
              }
            }
          ]
        }
      ]
    }
```
</details>

src/utils/content-to-jsx.tsx
fred-atherden: Yup, thanks. That LGTM 👍

fred-atherden:
@discodavey, they can both be there at the same time.

discodavey:
@scottaubrey  and @fred-atherden Does this look ok?

scottaubrey (approved):
This looks like it would be fine.

At the risk of saying "one more thing...", but it's hard to test this without an example in storybook. Would you find that helpful here, or as a follow up @discodavey ?

fred-atherden (approved):

src/utils/content-to-jsx.tsx
```diff
@@ -52,6 +50,15 @@ export const contentToJsx = (content: Content, index?: number): JSXContent => {
       return <li key={index}>{contentToJsx(content.content)}</li>;
     case 'List':
       return content.order === 'Ascending' ? <ol key={index}>{contentToJsx(content.items)}</ol> : <ul key={index}>{contentToJsx(content.items)}</ul>;
+    case 'Claim':
+      return (
+        <section key={index}>
+          {(content.label || content.title) &&
+            <h4>{content.label && contentToJsx(content.label)} {content.title && contentToJsx(content.title)}</h4>
```
E-Kemp, line 57: If there is just title will there be a leading space?

discodavey: @E-Kemp Need some of your expert knowledge on React

discodavey:
Thinking about this over the weekend and it's the space which is stopping this from going live. This is very similar to the social icons issue from the other week so this can go live and do a separate ticket for the space issue separately.

Any objections from anyone

scottaubrey:
> so this can go live and do a separate ticket for the space issue separately.
> 
> Any objections from anyone

While the time pressure isn't there this time, I think this is a fine to go as I think we're not even displaying content now even if it's there. The fix for spacing in the event of a title but no label (which in reality will affect no readers as the space isn't rendered) and adding an example in storybook I think can follow up.

PR #572, opened 2023-04-11
fix site width to 1200px

discodavey:
This makes sure site is max 1296px with padding


discodavey:
@chris-huggins Please can you look at the preview and let me know if this is ok. max-width of the site is now 1200 with 48px padding either side.

chris-huggins:
This looks great to me, thanks @discodavey

scottaubrey (approved):
I don't understand it, but it looks better and the number makes sense

PR #579, opened 2023-04-13
refactor so jump menu does not appear if no id

discodavey:


scottaubrey:
Actually - 1 last comment: we should probably add a test for this behaviour

scottaubrey (changes requested):
We should probably create a test for this.

discodavey:
@scottaubrey Added test

scottaubrey (approved):

PR #581, opened 2023-04-13
Restructure the data folder to support multiple versions and use path rather than doi

nlisgo:


src/components/pages/article/article-page.tsx
```diff
@@ -31,18 +32,19 @@ export type ArticlePageProps = {
 
 export const ArticlePage = (props: ArticlePageProps): JSX.Element => {
   const [activeTab, setActiveTab] = useState<string>(props.activeTab);
+  const id = props.msidWithVersion ?? props.metaData.msid;
```
nlisgo, line 35: Not requiring `msidWithVersion` was really an approach that I suspected would allow for less refactoring on the tests and we should probably make it a required property and refactor the tests.

`msidWithVersion` in it's current implementation can be one of 3 variations:

- [msid] (for published latest)
- [msid]v[version] (for published version specific)
- [msid]/v[version] (for previews)

These have more to do with the path and so `msidWithVersion` might not be the best name. We could call it `routeId`. I will refactor this and drop the `props.metaData.msid` fallback and then we will deal with the refactors to the tests.

data/manuscripts.json
scottaubrey, line 1: Couldn't think of a better place to put this @nlisgo - It worth pointing out that the "articleType" here would need to read "Reviewed Preprint" for one version, and "Revised Preprint" for another. We might need to restructure this to move other properties to the version (status, timeline maybe (to show "this version" or "go to version" links for different versions). This is where the negotiation about how far to go could come in.

docker-compose.yaml
```diff
@@ -74,11 +74,12 @@ services:
       /bin/sh -c "
       /usr/bin/mc config host add myminio http://minio:9000 minio miniotest;
       /usr/bin/mc mb myminio/eppdata;
+      /usr/bin/mc anonymous set public myminio/eppdata;
```
scottaubrey, line 77: Is this how we intend to do it in production, as opposed to using presigned URLs or a proxy?

Feels a bit blunt and open, but if we're not worried about that then I suppose it's OK. I just assumed we would want some control/audit over the requests made to the bucket.

nlisgo: Have removed now. Since: https://github.com/elifesciences/enhanced-preprints-client/pull/584

src/pages/api/reviewed-preprints.page.ts
```diff
@@ -19,7 +19,7 @@ type ReviewedPreprintItemResponse = {
 type ReviewedPreprintSnippet = {
   id: string,
   doi: string,
-  pdf: string,
+  pdf?: string,
```
scottaubrey, line 22: @nlisgo This is/was a little concerning - do you foresee any consequences of this for API / Journal?

scottaubrey: Thanks, glad it was just marked wrong here 👍

src/components/pages/article/article-page.tsx
scottaubrey: Yes, I think we're begging to be more explicit with these IDs. Proposal:
msid: 12345
doi: 10.7554/12345
versionId(entifier): 2
versionDoi: 10.7554/12345.2
routeId: 12345v2
contentId: 12345/v2

scottaubrey (approved):
Based on the conversation yesterday - I'm OK with merging this, then working on the extra entries and config for 85111v2

src/components/pages/article/article-page.tsx
nlisgo: Let's create a ticket with the above

nlisgo: Created: https://github.com/elifesciences/enhanced-preprints-issues/issues/621

PR #588, opened 2023-04-19
Storybook 7 upgrade

scottaubrey:
Upgrades to Storybook 7
Removes some workarounds that are [now handled by core](https://storybook.js.org/blog/integrate-nextjs-and-storybook-automatically/).
Slight moving around to still workaround next/image in tests.

I haven't done much more than running migrations, making sure it starts up and clicking around to make sure it seems to still work + fixing up the failing tests. Oh, and removing the reference to create-react-app.


will-byrne (approved):

scottaubrey:
> Approved but I would like to spend an afternoon at some point going over the stories to see if we can use the new version better. The docs section looks like it could be quite useful if we use it properly.

Agreed

PR #591, opened 2023-04-21
Manuscript aliases

scottaubrey:
Small feature that should make it easier to maintain multiple versions. Simple alias the non-versioned routeId to another routeId in manuscripts.json, e.g.

```
"80494": "80494v1",
```

The wiring up in manuscripts.ts will keep fetching the content for that key (recursively) until it's no longer a string. It also removes entries that are wrong, either by pointing to entries that don't exist, or by having a loop.


nlisgo (approved):
LGTM. merging if green

PR #603, opened 2023-04-24
Timeline aria labels

discodavey:


discodavey:
Open to suggestions on how to improve this word wise but this is the best we have at the moment with what is available to me

src/components/molecules/timeline/timeline.tsx
```diff
@@ -18,14 +18,14 @@ const formatDate = (date: string): string => new Date(date).toLocaleString('en-G
...
-    <dl className="review-timeline__list">
+    <dl className="review-timeline__list" aria-label="This is the timeline of this Reviewed Preprint">
       {
         events.sort((a, b) => new Date(b.date).getTime() - new Date(a.date).getTime()).map((entry, index) => (
           <Fragment key={index}>
             <dt className="review-timeline__event">{entry.name}</dt>
             <dd className="review-timeline__date">
               {formatDate(entry.date)}
-              { entry.link && <a className="review-timeline__link" href={entry.link.url}>{entry.link.text}</a>}
+              { entry.link && <a className="review-timeline__link" href={entry.link.url} aria-label={entry.name}>{entry.link.text}</a>}
```
nlisgo, line 28: Shall we add a test in for the `aria-label`? Also, in terms of improvements I would ask if this allows us to add the `aria-label` we want? For the examples we have in storybook are you happy with using the name, if not then we could allow it to be set by a new variable and perhaps use the name as fallback?

discodavey: Would there be a time where we would need to put in a custom aria-label name as we will be using {entry.name} at this time? It should be the name as it will then say what the link is for unless you know of any other cases where this would change.

Do we need to complicate it at this moment?

We can put a test in for the aria-label but this would need to be optional as not all of the timeline.

E-Kemp: I don't see a reason to overcomplicate the aria-label, however there is no existing test for the conditional links which we should probably add

E-Kemp (approved):

src/components/molecules/timeline/timeline.tsx
```diff
@@ -18,14 +18,14 @@ const formatDate = (date: string): string => new Date(date).toLocaleString('en-G
 
 export const Timeline = ({ events }: TimelineProps): JSX.Element => (
   <div className="review-timeline">
-    <dl className="review-timeline__list">
+    <dl className="review-timeline__list" aria-label="This is the timeline of this Reviewed Preprint">
```
nlisgo, line 21: As discussed, let's make this `aria-label` value a parameter, could be required. We can add an example in storybook.

E-Kemp (approved):
Looks good to me, nice detail in the tests! 👍

PR #631, opened 2023-04-27
Google scholar post validation

discodavey:
relates to https://github.com/elifesciences/enhanced-preprints-issues/issues/631


scottaubrey (changes requested):
Could we change the doi too please. I think we might need to construct from other data like:
```
const doi = getRppDoi(props.metaData);
```

discodavey:
@fred-atherden Is this the correct id I need to be using for this?

```<meta name="citation_id" content={props.metaData.msid}/>```

fred-atherden:
Hi @discodavey, thanks for looping me in. I have an answer and a few points/questions based on this PR:

1. I think `citation_id` is equivalent to the eLocation Id in the citation so it needs to be `RP${props.metaData.msid}`, i.e. for 86464 it should be `RP86464`. 

2. If I've understood correctly (based on [here](https://github.com/elifesciences/enhanced-preprints-client/blob/0f38d4e085daa9d38544fbf5822d943d14972208/src/manuscripts.ts#L71)), it looks like the version DOI is being included for `citation_doi`. I think we need to include the umbrella DOI for Google Scholar (i.e. the one without the version suffix). Will that be possible?

3. Would `metaData.publishedYear` change when a new version is published or remain static? The reason I ask is:
a. It looks like it's being used to determine the volume (which is correct, although the volume needs to stay the same - the first version publication date - 2011).
b. If I've understood correctly `citation_publication_date` is derived from elsewhere (`props.status.timeline`)? And I'm wondering why.

4. The value of `content` for [citation_language](https://github.com/elifesciences/enhanced-preprints-client/blob/0f38d4e085daa9d38544fbf5822d943d14972208/src/pages/reviewed-preprints/%5Bmsid%5D/index.page.tsx#L44) should be `en` not `text/html` - this declares the language the content is written in, which for eLife content (at the moment) will always be english. 

5. For this line https://github.com/elifesciences/enhanced-preprints-client/blob/0f38d4e085daa9d38544fbf5822d943d14972208/src/pages/reviewed-preprints/%5Bmsid%5D/index.page.tsx#L45 there [was a recent PR by Elliot](https://github.com/elifesciences/enhanced-preprints-client/pull/636) which handled authors (in references) that don't have either `givenNames` or `familyNames` (because the optional chaining in that expression introduced the text `undefined` in place of `givenNames` or `familyNames` when not present). It might be worth incorporating that change here too. 

There's an example of an RP with an author that only has `familyNames` [here](https://elifesciences.org/reviewed-preprints/84385) - the last one.

discodavey:
@fred-atherden In answer to your questions

3. We are keeping publication year separate so that the volume does not change with new versions.

Rest of your points have been done in the code.

Can you please check and approve

scottaubrey (approved):

fred-atherden (approved):

PR #636, opened 2023-04-28
Fix given names being possibly undefined

E-Kemp:


nlisgo (approved):

fred-atherden:
@nlisgo, @E-Kemp many thanks for this - can we do the same thing for the citation modal?

Here's an example from https://elifesciences.org/reviewed-preprints/84385:
Screenshot 2023-05-02 at 09 04 07

@QueenKraken I assume this would need to be raised as a ticket? It's not urgent but would be good to fix at some point in the near future.

QueenKraken:
Hi @fred-atherden yes we can - please raise the request through the engagement channel so we can track it.

PR #639, opened 2023-04-28
Shallow preload pages and lazy load images

scottaubrey:


src/pages/reviewed-preprints/[msid]/[...tab].page.tsx
```diff
@@ -27,24 +28,38 @@ const getPublishedDate = (events: TimelineEvent[]): string | undefined => {
...
-    <meta name="citation_language" content="text/html"/>
-    { props.metaData.authors.map((author, index) => <meta key={index} name="citation_author" content={`${author.givenNames?.join(' ')} ${author.familyNames?.join(' ')}`}/>)}
-  </Head>
-  <ArticlePage metaData={props.metaData} msidWithVersion={props.msidWithVersion} status={props.status} activeTab="fulltext">
-    <ArticleFullTextTab content={props.content} metaData={props.metaData} peerReview={props.peerReview}></ArticleFullTextTab>
-  </ArticlePage>
-  </>
-);
+export const Page = (props: PageProps): JSX.Element => {
+  const tabs: { [key: string]: () => JSX.Element } = {
```
Comment (E-Kemp) [inline src/pages/reviewed-preprints/[msid]/[...tab].page.tsx:32]: I'm not sure if/how Next would export this into client side JavaScript, maybe this is worth lifting out of the function component? (just a couple of lines above)

src/pages/reviewed-preprints/[msid]/[...tab].page.tsx
```diff
@@ -27,24 +28,38 @@ const getPublishedDate = (events: TimelineEvent[]): string | undefined => {
...
+    fulltext: () => <ArticleFullTextTab content={props.content} metaData={props.metaData} peerReview={props.peerReview}></ArticleFullTextTab>,
+    figures: () => <ArticleFiguresTab content={props.content}></ArticleFiguresTab>,
+    reviews: () => <ArticleReviewsTab peerReview={props.peerReview}></ArticleReviewsTab>,
+    pdf: () => (<>
+        <ArticleFullTextTab content={props.content} metaData={props.metaData} peerReview={props.peerReview}></ArticleFullTextTab>
+        <ArticleReviewsTab peerReview={props.peerReview}></ArticleReviewsTab>
+      </>),
+  };
+  const router = useRouter();
+  const tabName = (router.query.tab?.[0] && tabs[router.query.tab[0]] !== undefined) ? router.query.tab[0] : 'fulltext';
```
Comment (E-Kemp) [inline src/pages/reviewed-preprints/[msid]/[...tab].page.tsx:42]: I'm wondering if this line and the line below would be better served as memos that change dependent on router.query.tab so they don't recalculate upon every render on the client (unless that's not how Next handles it?)

E-Kemp (approved):
Ultimately I'm really happy with this, the result is identical to my original PR but it takes advantage of some of NextJS's best tools; I'm happy for it to go instead of my client-oriented PR

(I'm surprised I didn't think of this honestly!)

nlisgo:
When we click on a tab for the first time then we should expect that we can see the top of the tabbed content. Currently, if you are on the fulltext tab and scroll to near the bottom of the page and then click on the figures or reviews tab then you will just be seeing the footer:

- Visit https://pr-639--epp.elifesciences.org/reviewed-preprints/80494v1#references
- Click on the "Figures and data" tab
- The expected behaviour would be that you can see the top of the figures content. @chris-huggins can clarify the ideal behaviour on his return but I would assume it would be acceptable to hold the position of the scroll if it is between 0 and the tabs reaching the top of the browser window, if scroll is beyond there then clicking on the tabs will result in the tabbed content being loaded and the scroll offset to be such that the tabs appear at the top of the browser window.

scottaubrey:
> When we click on a tab for the first time then we should expect that we can see the top of the tabbed content. Currently, if you are on the fulltext tab and scroll to near the bottom of the page and then click on the figures or reviews tab then you will just be seeing the footer:
> 
> * Visit https://pr-639--epp.elifesciences.org/reviewed-preprints/80494v1#references
> * Click on the "Figures and data" tab
> * The expected behaviour would be that you can see the top of the figures content. @chris-huggins can clarify the ideal behaviour on his return but I would assume it would be acceptable to hold the position of the scroll if it is between 0 and the tabs reaching the top of the browser window, if scroll is beyond there then clicking on the tabs will result in the tabbed content being loaded and the scroll offset to be such that the tabs appear at the top of the browser window.

Ideally, I think jumping to content would be a good improvement like you suggest above (and adding the anchor would be good for accessibility too), but the behaviour of the smooth scrolling with jumping to content between tabs is weird. We'd need to address that wider interactions, and possibly only add to links once scrolled past (which feels not difficult, but easy to get wrong in a bad way), or manually scroll to content - all of which feels like an improvement in itself. 

For these reasons, I've just recreated the current behaviour of jumping to the top of the page on page load / tab switch.

src/pages/reviewed-preprints/[msid]/[...tab].page.tsx
Comment (scottaubrey) [reply src/pages/reviewed-preprints/[msid]/[...tab].page.tsx:33]: This was just to make the typescript compiler happy - delivered to the client it's removed during the compile step (as all types are). I'm happy to move it, but I'm not sure it's adding much as a standalone type.

src/pages/reviewed-preprints/[msid]/[...tab].page.tsx
Comment (scottaubrey) [reply src/pages/reviewed-preprints/[msid]/[...tab].page.tsx]: Adding a console log to track when it's trigger, it looks to be only on page load that that page component is called.

Having said that - if that was to change there's no harm in adding the `useMemo` and it seems to work as we expect around `router.query.tab` (or at least triggers as often as it ever did). Added! 👍

nlisgo (approved):

PR #666, opened 2023-05-05
Allow for an event description without link

nlisgo:


src/components/molecules/timeline/timeline.stories.tsx
```diff
@@ -40,6 +40,38 @@ EventTimelineWithLinks.args = {
...
+      },
+    },
+    {
+      name: 'event 2',
+      date: '2000-02-02',
+      link: {
+        text: 'Go to version',
+        url: 'https://version.url',
+      },
+      eventDescription: 'Version description',
```
nlisgo, line 61: When `eventDescription` is present along with link then the `eventDescription` is used as the aria label.

nlisgo:
You can preview in storybook here: https://pr-666--epp-storybook.elifesciences.org/?path=/story/molecules-timeline--event-timeline-with-descriptions

scottaubrey (dismissed):
Looks OK to me, though the interplay between event description and link wouldn't have been my preference - I prefer explicit over implicit. I don't think it will matter though.

src/components/molecules/timeline/timeline.tsx
```diff
@@ -1,17 +1,33 @@
...
+  },
+  eventDescription?: undefined,
 };
 
+type TimelineEventWithDescription = TimelineEventBasic & {
+  link?: undefined,
+  eventDescription: string,
+};
+
+export type TimelineEvent = TimelineEventSimple | TimelineEventWithLink | TimelineEventWithDescription;
```
nlisgo, line 27: I had initially approached it as follows but couldn't get it to work:

```
type TimelineEventBasic = {
  name: string,
  date: string,
};

type TimelineEventWithLink = TimelineEventBasic & {
  link: {
    text: string,
    url?: string,
  },
};

type TimelineEventWithDescription = TimelineEventBasic & {
  eventDescription: string,
};

export type TimelineEvent = TimelineEventBasic | TimelineEventWithLink | TimelineEventWithDescription;
```

nlisgo:
Thanks @scottaubrey for the review but I was a little premature in asking for one. This now includes the modifications to `manuscripts.json` and the modification to the script to alter the `manuscripts.json`.

Your feedback was valuable as I wasn't entirely happy with `eventDescription`. I've opted since your review to allow `link` or `eventDescription` but not both.

nlisgo:
@chris-huggins would appreciate your review/feedback: https://pr-666--epp.elifesciences.org/reviewed-preprints/86784

chris-huggins:
Thanks @nlisgo. I'm adding feedback here for reference but happy to explain and discuss options on a call if/when you're available.

If I compare the PR demo with designs it seems the "This version" text is different to how I'd intended, which was to use brackets in place of the `•` dot separator to intentionally make it different to the links to versions/preprint server. The demo doesn't have a "Go to version" link but worth considering what this would look like too.

**PR Demo** (worth noting there is no hover state on the Go to bioRxiv link. "This version" is not a link)

image


**Design:** (showing "this version" in brackets, lower case, and without `•`)

image


As it is currently, the `Go to bioRxiv` link and `This version` text are styled the same, with only noticeable difference the cursor change on hover. I don't think this is sufficient so I think we should discuss if we can change `This version` to match the designs, or change how we style/indicate the `Go to bioRxiv` link (with `Go to version` link styling in mind too).

The links have poor affordance anyway (they're deliberately subtle) so I'd be happy to look at ways to change the link styling if it is harder to change "This version" text to match the designs.

chris-huggins:
Thanks @nlisgo. I'm adding feedback here for reference but happy to explain and discuss options on a call if/when you're available.

If I compare the PR demo with designs it seems the "This version" text is different to how I'd intended, which was to use brackets in place of the `•` dot separator to intentionally make it different to the links to versions/preprint server. The demo doesn't have a "Go to version" link but worth considering what this would look like too.

**PR Demo** (worth noting there is no hover state on the Go to bioRxiv link. "This version" is not a link)

image


**Design:** (showing "this version" in brackets, lower case, and without `•`)

image


As it is currently, the `Go to bioRxiv` link and `This version` text are styled the same, with only noticeable difference the cursor change on hover. I don't think this is sufficient so I think we should discuss if we can change `This version` to match the designs, or change how we style/indicate the `Go to bioRxiv` link (with `Go to version` link styling in mind too).

The links have poor affordance anyway (they're deliberately subtle) so I'd be happy to look at ways to change the link styling if it is harder to change "This version" text to match the designs.

chris-huggins:
Thanks @nlisgo. I'm adding feedback here for reference but happy to explain and discuss options on a call if/when you're available.

If I compare the PR demo with designs it seems the "This version" text is different to how I'd intended, which was to use brackets in place of the `•` dot separator to intentionally make it different to the links to versions/preprint server. The demo doesn't have a "Go to version" link but worth considering what this would look like too.

**PR Demo** (worth noting there is no hover state on the Go to bioRxiv link. "This version" is not a link)

image


**Design:** (showing "this version" in brackets, lower case, and without `•`)

image


As it is currently, the `Go to bioRxiv` link and `This version` text are styled the same, with only noticeable difference the cursor change on hover. I don't think this is sufficient so I think we should discuss if we can change `This version` to match the designs, or change how we style/indicate the `Go to bioRxiv` link (with `Go to version` link styling in mind too).

The links have poor affordance anyway (they're deliberately subtle) so I'd be happy to look at ways to change the link styling if it is harder to change "This version" text to match the designs.

nlisgo:
@chris-huggins as discussed the updated preview has the text change to `(this version)` and I have removed the bullet.

Let me know what you think.

chris-huggins:
It looks great to me, thanks @nlisgo.

src/components/molecules/timeline/timeline.tsx
scottaubrey: [Because](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#working-with-union-types):
> TypeScript will only allow an operation if it is valid for every member of the union. For example, if you have the union string | number, you can’t use methods that are only available on string:

The way around it is with typeGuard functions:
```
const isTimelineEventWithLink = (event: TimelineEvent): event is TimelineEventWithLink => (event as TimelineEventWithLink).link === undefined;
const isTimelineEventWithDescription = (event: TimelineEvent): event is TimelineEventWithDescription => (event as TimelineEventWithDescription).eventDescription === undefined;
const isTimelineEventBasic = (event: TimelineEvent): event is TimelineEventBasic => !isTimelineEventWithLink(event) && !isTimelineEventWithDescription(event);
```

then you could change the code like so:

```
diff --git a/src/components/molecules/timeline/timeline.tsx b/src/components/molecules/timeline/timeline.tsx
index 046c1eb..734adae 100644
--- a/src/components/molecules/timeline/timeline.tsx
+++ b/src/components/molecules/timeline/timeline.tsx
@@ -6,25 +6,22 @@ type TimelineEventBasic = {
   date: string,
 };
 
-type TimelineEventSimple = TimelineEventBasic & {
-  link?: undefined,
-  eventDescription?: undefined,
-};
-
 type TimelineEventWithLink = TimelineEventBasic & {
   link: {
     text: string,
     url?: string,
   },
-  eventDescription?: undefined,
 };
 
 type TimelineEventWithDescription = TimelineEventBasic & {
-  link?: undefined,
   eventDescription: string,
 };
 
-export type TimelineEvent = TimelineEventSimple | TimelineEventWithLink | TimelineEventWithDescription;
+export type TimelineEvent = TimelineEventBasic | TimelineEventWithLink | TimelineEventWithDescription;
+
+const isTimelineEventWithLink = (event: TimelineEvent): event is TimelineEventWithLink => (event as TimelineEventWithLink).link === undefined;
+const isTimelineEventWithDescription = (event: TimelineEvent): event is TimelineEventWithDescription => (event as TimelineEventWithDescription).eventDescription === undefined;
+// const isTimelineEventBasic = (event: TimelineEvent): event is TimelineEventBasic => !isTimelineEventWithLink(event) && !isTimelineEventWithDescription(event);
 
 type TimelineProps = {
   events: Array<TimelineEvent>,
@@ -42,8 +39,8 @@ export const Timeline = ({ events, listDescription }: TimelineProps): JSX.Elemen
             <dt className="review-timeline__event">{entry.name}</dt>
             <dd className="review-timeline__date">
               {formatDate(entry.date)}
-              {entry.link && <a className="review-timeline__link" href={entry.link.url} aria-label={entry.name}>{entry.link.text}</a>}
-              {entry.eventDescription && <span className="review-timeline__description"> {entry.eventDescription}</span>}
+              {isTimelineEventWithLink(entry) && <a className="review-timeline__link" href={entry.link.url} aria-label={entry.name}>{entry.link.text}</a>}
+              {isTimelineEventWithDescription(entry) && <span className="review-timeline__description"> {entry.eventDescription}</span>}
             </dd>
           </Fragment>
         ))
```

I think this is the cleaner way to express your original types, but at the cost of a few extra functions for type guarding.

scottaubrey (approved):

PR #679, opened 2023-05-09
Introduce hover color for timeline links

nlisgo:


nlisgo:
@chris-huggins when the preview is available please verify that this is working as expected. Thanks.

scottaubrey (approved):

PR #685, opened 2023-05-10
Pass id through for reviews

nlisgo:
This can be considered for review after the following PR has been deployed to production: https://github.com/elifesciences/enhanced-preprints-server/pull/690


nlisgo:
Now the corresponding server PR has been merged in I will mark this ready for review once the server code has been deployed to production.

scottaubrey (approved):
This looks good, nice and simple. 

I've not run the code, but the preview env works as expected against staging API, so I'm happy to go with this. 👍

scottaubrey:
> This looks good, nice and simple.
> 
> I've not run the code, but the preview env works as expected against staging API, so I'm happy to go with this. 👍

@nlisgo One final thing - the reviews.json override in the cluster also needs to be undone too I assume?

nlisgo:
> @nlisgo One final thing - the reviews.json override in the cluster also needs to be undone too I assume?

The plan would be to deploy this change and then when it's in production demonstrate that removing the reviews for 85111 will result in no change, then remove it.

PR #700, opened 2023-05-15
Dark mode sandbox

discodavey:


discodavey:
@chris-huggins This is as much as I have done for dark mode so far. You can have a look but I have just done the basics.

If you can come up with what you want in terms of colours, supply the svg's and images then I can make the changes.

https://pr-700--epp.elifesciences.org/

Not done anything with Storybook at the moment

PR #709, opened 2023-05-16
Fix link colour for assessments section

discodavey:


discodavey:
@chris-huggins Can you check you are happy. Just a colour change

chris-huggins:
Looks good to me, thanks @discodavey

chris-huggins (approved):

PR #723, opened 2023-05-18
Adjust workflows so that manuscripts.json is modified in 2 steps

nlisgo:


.github/workflows/add-pdf-url-to-manuscript.yaml
```diff
@@ -32,18 +32,22 @@ jobs:
...
-            tee manuscripts.json > /dev/null
+            tee manuscripts-new.json > /dev/null
+      - name: Replace manuscripts.json
+        run: |
+          mv manuscripts-new.json manuscripts.json
       - name: Create Pull Request (master)
         if: github.ref == 'refs/heads/master'
         uses: peter-evans/create-pull-request@v4
         with:
+          draft: true
```
nlisgo, line 50: This will create a draft PR initially so that it's easier to trigger CI.

.github/workflows/add-pdf-url-to-manuscript.yaml
```diff
@@ -32,18 +32,22 @@ jobs:
...
-      - name: Add PDF url to manuscript
+      - name: Prepare manuscripts.json with PDF url
         run: |
           cat manuscripts.json | \
             yarn --silent add-pdf-url-to-manuscript \
             --msid "${{ github.event.inputs.msid }}" \
             --versionManuscript "${{ github.event.inputs.versionManuscript || 1 }}" \
             --urlPdf "${{ github.event.inputs.urlPdf }}" | \
-            tee manuscripts.json > /dev/null
+            tee manuscripts-new.json > /dev/null
```
nlisgo, line 42: I think the issue was caused by beginning to write to the output stream of the same file that we were reading in. This is working for me.

fred-atherden (approved):
I have tested locally and it works for me - thanks!

PR #763, opened 2023-06-01
Fix preprint posted date for 85795

fred-atherden:


fred-atherden:
@nlisgo, are you happy with this manual approach to fix an incorrect date in the timeline for 85795 (introduced in [c9a924e](https://github.com/elifesciences/enhanced-preprints-client/pull/618/commits/c9a924ebe8482a46b515149a579471ed666d46e3)) ?

I assume there's no way to change this via the existing workflows, and it seems unnecessary to create one for this purpose.

nlisgo (approved):

PR #780, opened 2023-06-05
Fixed middle tab button styling

E-Kemp:
Closes https://github.com/elifesciences/enhanced-preprints-issues/issues/653


src/components/pages/article/article-page.scss
```diff
@@ -100,11 +104,11 @@
...
   }
 
-  @media only screen and (max-width: settings.$breakpoint-small) {
+  @media only screen and (min-width: settings.$breakpoint-small) {
     &:nth-child(2) {
-      overflow: hidden;
-      white-space: nowrap;
-      width: 52px;
+      ::after {
+        content: " and data";
```
discodavey, line 110: Not a fan of doing this as it means all users do not get the same experience

E-Kemp: Is this a question of browser compatibility or the different text giving a different experience? It's reflecting how it looks in the designs, maybe one for @chris-huggins to look at?

chris-huggins: For the record I've spoken with dave about this and we can talk in future about if the design needs to change to include "And data" at all screen sizes. But in the meantime this PR can go ahead to fix the issue with "Figures" being cut off.

chris-huggins (approved):

PR #813, opened 2023-06-14
Add additional nextJS linting plugin

scottaubrey:
```
- warn The Next.js plugin was not detected in your ESLint configuration. See https://nextjs.org/docs/basic-features/eslint#migrating-existing-config
```

This PR "fixes" this.

It also fixes the one linting fail that resulted from this (though there are a few warnings too)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,7 +5,7 @@ module.exports = {
     project: './tsconfig.json',
   },
   plugins: ['@typescript-eslint', "no-only-tests"],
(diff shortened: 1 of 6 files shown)


E-Kemp (changes requested):
There's some check annotations as a result of linting in the test files, could this be comment skipped before we merge? Otherwise I think it could come up on every future PR

scottaubrey:
> There's some check annotations as a result of linting in the test files (look at the bottom of [here](https://github.com/elifesciences/enhanced-preprints-client/pull/813/files)) could this be comment skipped before we merge? Otherwise I think it could come up on every future PR

They are warnings, so don't stop anything progressing, and I'm uncertain if we want to move to next/Image for a component like this - perhaps a skip comment is worthwhile. 

I'll add it.

E-Kemp (approved):

PR #887, opened 2023-07-04
Control showing previews

scottaubrey:
This branch shows two ways we could support previews clientside:
- env var
- via URL

the url route is a little bit fiddly - configure a forwarding route, read , but now we can handle a review-less article it does allow us now to reuse most of the same code (unlike the old preview path). This will mean preexisting `/preview` page controller can be deleted after automation is live.


nlisgo:
Postponing review until after the following PR is merged in as there seems to be a lot of shared code: https://github.com/elifesciences/enhanced-preprints-client/pull/883

scottaubrey:
> Postponing review until after the following PR is merged in as there seems to be a lot of shared code: #883

It's merged now and integrated here.

This PR isn't necessarily to be merged (though it could), but more for a conversation and discussion if this is a right way following our conversation @nlisgo 

This is certainly a very simple way to handle preview/published. However, it will look published in the docmap and EPP APIs. However, all the information in docmaps and thus the EPP API are all public by that point anyway, so I'm still unclear if we need to/should be taking embargoing that seriously or not. Might be a conversation for @QueenKraken to be part of too.

Anyway, the alternatives as I see it:
- proxy it through a client (next) API that filters out if previews are allowed, don't expose the server API publicly anymore.
- set a var at the env level in server/API to show/not show previews in staging/prod:
- build the same `isPreview` logic in at the server API level, 
   - either additional flag to request previews, 
   - or entirely different API (`/api/previews/{msid}`)

nlisgo:
@scottaubrey let's chat about this one tomorrow.

will-byrne (approved):

PR #894, opened 2023-07-06
Remove brackets from single Cite block

nlisgo:
Needed before we upgrade to Encoda 1.0.0 but we should release shortly after upgrade.


nlisgo:
Let's discuss impact of this tomorrow morning.

scottaubrey (approved):

PR #917, opened 2023-07-12
Fixes for new preview tabs being incorrect, misc fixes for article page types

scottaubrey:


src/pages/reviewed-preprints/[...path].page.tsx
```diff
@@ -68,8 +69,8 @@ export const Page = (props: PageProps): JSX.Element => {
     pdf: {
       tabLinks: [],
       content: () => (<>
-        <ArticleFullTextTab content={props.content} metaData={props.metaData} peerReview={props.peerReview ?? undefined}></ArticleFullTextTab>
```
Comment (scottaubrey) [inline src/pages/reviewed-preprints/[...path].page.tsx:71]: This is just repeated code, so I refactored rather than make the change twice.

src/pages/reviewed-preprints/[...path].page.tsx
```diff
@@ -186,7 +188,10 @@ export const getServerSideProps: GetServerSideProps<PageProps> = async (context:
       },
       msidWithVersion: id,
       content,
-      status,
+      status: {
+        ...status,
+        isPreview: false,
```
Comment (scottaubrey) [inline src/pages/reviewed-preprints/[...path].page.tsx:193]: when in non-automated mode, this page is always only published manuscripts

src/pages/preview/[...path].page.tsx
```diff
@@ -80,6 +88,7 @@ export const getServerSideProps: GetServerSideProps = async (context: GetServerS
         timeline: [
           { name: 'Preview Preprint generated', date: new Date().toDateString() },
         ],
+        isPreview: true,
```
Comment (scottaubrey) [inline src/pages/preview/[...path].page.tsx:91]: this is always a preview

src/pages/preview/[...path].page.tsx
Comment (scottaubrey) [inline src/pages/preview/[...path].page.tsx:1]: Quite a lot of this change was actually because the current preview page is not passing required page props. So it was sort of working, because eventually the prop is passed to an optional, but as soon as I added the generic param to `getServerSideProps` it all went mental. I also imported some of the new tab-selection logic from the manuscript page, as there were bugs with this old logic, and it made it hard to unify the two PageProps.

src/components/pages/article/tabs/fulltext-tab.test.tsx
```diff
@@ -7,9 +7,21 @@ describe('FulltextTab', () => {
     expect(() => render(<ArticleFullTextTab content={content} metaData={metaData} peerReview={peerReview}/>)).not.toThrow();
   });
 
+  it('does not render the evaluation summary', () => {
+    const { container } = render(<ArticleFullTextTab content={content} metaData={metaData}/>);
+
+    expect(container.querySelector('#assessment')).toBeNull();
+  });
+
   it('renders the evaluation summary when one is passed in', () => {
```
scottaubrey, line 16: This test started failing, but I took the opportunity to add more tests too

nlisgo (approved):

PR #966, opened 2023-07-27
feat: Add Node.js debugging capabilities to Docker and Next.js dev script

nlisgo:


package.json
```diff
@@ -14,7 +14,7 @@
   "scripts": {
     "build": "next build",
     "start": "next start",
-    "start:dev": "next dev",
+    "start:dev": "NODE_OPTIONS='--inspect' next dev",
```
scottaubrey, line 17: I had wondered about creating a `start:debug` that has a value `"NODE_OPTIONS='--inspect-brk' next dev"`, so that node waits for a debugger to connect before running the main script.

scottaubrey (commented):
Were you able to add break points? I was struggling to get the debugger to recognise them without setting `"sourceMap": true,` in tsconfig.ts and running locally. I never got it working in the container.

package.json
nlisgo: Whichever one we use I would like it to use it with docker so that we can have server working with it. At the moment start:dev is triggered in the Dockerfile.

Will merge this in and we can adapt it as need requires.

nlisgo: I tried `--inspect-brk` with docker and app was unhealthy.

scottaubrey: The point of it would be to stop until the debugger connects - are you saying that you connected and it still didn't proceed? That sounds more like you didn't connect to the "right" node process ?

PR #1013, opened 2023-08-09
fix: remove default value for background_image in button-icon mixin

nlisgo:
if we wish to allow a default value of none it should not use the url()

see: https://stackoverflow.com/questions/1161061/css-background-image-urlnone-leading-to-errors-in-apache-log-file

### DIFF
--- a/src/components/atoms/button/button.scss
+++ b/src/components/atoms/button/button.scss
@@ -56,7 +56,7 @@
   }
 }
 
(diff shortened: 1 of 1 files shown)


nlisgo:
Discussed with @viktoria-mahmud and we are both happy for this to go in.

cc @discodavey (tagged as it is an adjustment to something you recently introduced, happy to discuss when you get back)

PR #1047, opened 2023-08-16
feat: update msas field to use article subjects

nlisgo:
The msas field in the reviewed preprints page is now populated with the article subjects. Additionally, a new field called subjects has been added to the enhanced-article type.


TheMaroonHatHacker (approved):
All seems to be working as intended and nothing has broke in the process.

PR #1136, opened 2023-09-11
feat: handle potential undefined author names

nlisgo:
Updated the author name formatting across several components to handle potential undefined givenNames and familyNames. This supports better rendering of author information and prevents potential errors in displaying author data.


QueenKraken:
Fred has confirmed all resolved.

PR #1239, opened 2023-10-06
feat: update ClaimContent type and mocks

nlisgo:
- Made claimType optional in ClaimContent type
- Added a generic 'Claim' and a 'Statement' claim to the mocks

### DIFF
--- a/src/types/content.ts
+++ b/src/types/content.ts
@@ -80,7 +80,7 @@ type ListContent = {
 
 type ClaimContent = DecoratedContent & {
   type: 'Claim',
-  claimType: 'Statement' | 'Theorem' | 'Lemma' | 'Proof' | 'Postulate' | 'Hypothesis' | 'Proposition' | 'Corollary',
+  claimType?: 'Statement' | 'Theorem' | 'Lemma' | 'Proof' | 'Postulate' | 'Hypothesis' | 'Proposition' | 'Corollary',
   label?: Content,
   title?: Content,
 };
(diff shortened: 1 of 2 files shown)


src/types/content.ts
```diff
@@ -80,7 +80,7 @@ type ListContent = {
 
 type ClaimContent = DecoratedContent & {
   type: 'Claim',
-  claimType: 'Statement' | 'Theorem' | 'Lemma' | 'Proof' | 'Postulate' | 'Hypothesis' | 'Proposition' | 'Corollary',
+  claimType?: 'Statement' | 'Theorem' | 'Lemma' | 'Proof' | 'Postulate' | 'Hypothesis' | 'Proposition' | 'Corollary',
```
nlisgo, line 83: No change is needed to the template because `claimType` is not used:

https://github.com/elifesciences/enhanced-preprints-client/blob/master/src/utils/content-to-jsx.tsx#L59-L67

nlisgo:
We should adapt the process so that we can explicitly indicate that a PR doesn't require coko approval. If we clearly know that no change is made to the presentation.

scottaubrey:
> We should adapt the process so that we can explicitly indicate that a PR doesn't require coko approval. If we clearly know that no change is made to the presentation.

From my original intro:
> The label can be added either by us (if we know there’s no HTML [added: change]) or after getting the go ahead from Coko (via a preview env probably), so it’s lightweight, but does add a layer of protection for us when trying to merge.

I'm OK with us adding the coko-approved label if we know it's not a coko thing. If you think another label is worthwhile, it's a small change I guess we could add, but I think I'd rather work on improving the tech so it's not needed anymore (the HTML improvements that Paul has asked for us to do) rather than minor improvements to this  - it's only meant to make us stop and think rather than be a record or anything.

PR #1257, opened 2023-10-13
feat: make prepareAuthor function exportable and use it in [msid].page.ts

nlisgo:
The `prepareAuthor` function from `reviewed-preprints.page.ts` is now exportable. It is imported and utilized in `reviewed-preprints/[msid].page.ts` to prepare the authors' metadata.


nlisgo:
This should not require Coko approval. We should merge once CI pipeline is green.

PR #1433, opened 2023-11-29
change 'Posted to bioRxiv' to 'Posted to preprint server'

smoqadam:


scottaubrey (approved):

scottaubrey:
Nice and easy :)

scottaubrey:
We should check with Coko that this doesn't affect anything, but it can still be merged as this is the automation system.

PR #1469, opened 2023-12-05
fix broken ts in test

will-byrne:
Closes https://github.com/elifesciences/enhanced-preprints-issues/issues/892


scottaubrey (approved):
Any thoughts why it compiled/ran fine?

will-byrne:
> Any thoughts why it compiled/ran fine?

I think its because we are using ts-jest so its interpreting it as it runs and ignores the error. if we built the tests and ran them with jest (JS) it would probably complain. To be honest I think the best way forward is to just fix these issues when we find them and rely on our IDEs when writing them.

PR #1472, opened 2023-12-06
Fixes for finding image info, stop crashing when we don't have it

scottaubrey:


smoqadam (approved):

viktoria-mahmud:
All of these pdf pages are working with the fix:
88480v2
89780v2
90378v1
91532v1
92254v1
91397v1

PR #1509, opened 2023-12-15
feat: update 'published' and 'reviewedDate' to use 'firstPublished'

nlisgo:
This commit updates the 'published' and 'reviewedDate' properties in the 'reviewedPreprintSnippet' function to use 'firstPublished' instead of 'published'. Also, it adds 'firstPublished' as a new property in the 'EnhancedArticleNoContent' type. Additionally, it changes the import path of 'Author' in the 'reviewed-preprints.page.test.ts' file.


nlisgo:
This shouldn't be deployed until after: https://github.com/elifesciences/enhanced-preprints-server/pull/923

PR #1515, opened 2023-12-18
Rewrite journal API to use summary and full content APIs

scottaubrey:
This restores an alternative path whereby we utilise the server summary api to generate a list, then slice the list, then fetch further content from the full content API.

Possibly, we could optimise further by creating a nocontentbyid API, to not pull down the full content for the journal api index


scottaubrey:
Actually, with the improvements and the lack of PDF, this is now slower than the current API. Closing.

PR #1554, opened 2024-01-04
feat: add function to format date without milliseconds

nlisgo:
This commit adds a function to convert Date objects into ISO strings without milliseconds. This function is then used to format the 'published', 'reviewedDate', 'versionDate', and 'statusDate' fields in the 'reviewedPreprintSnippet' function.


nlisgo:
We should followup with a test but as this is causing issues I'm keen to get this live.

PR #1614, opened 2024-01-23
Remove superflous volume mounts that stop the browser tests running without having run the full suite beforehand

scottaubrey:
With a completely clean docker, you bring up the browser test setup by running `docker compose -f docker-compose.browsertest.yaml up` on it's own. It depends on a volume being filled by running `docker compose up` first.

This removes that dependency. Also separates it to another job which reduces overall CI time by about 45s


scottaubrey:
FYI @elifesciences/epp-tech-team

PR #1868, opened 2024-04-17
[#1045] Improve test and for pluralization of metrics

nlisgo:


scottaubrey:
Nice catch on the bug.

Since we've resisted introducing libraries into the rendering stack before due to it necessarily increasing payload size, I thought it worth mentioning that this seems to have added about 2kb to page delivery. In context, our page delivery size for 85111 imported in dev env is 552kb (or 550kb). Nothing noticable really in the grand schema of things.

nlisgo:
> Nice catch on the bug.
> 
> Since we've resisted introducing libraries into the rendering stack before due to it necessarily increasing payload size, I thought it worth mentioning that this seems to have added about 2kb to page delivery. In context, our page delivery size for 85111 imported in dev env is 552kb (or 550kb). Nothing noticable really in the grand schema of things.

We should discuss the approach of resisting introducing libraries. Is there a way to minimise or offset the use of libraries. It would be great even for small utility functions to support light libraries. It obviously would be very easy to write this function but I believed I was doing the right thing by using a library. Are we able to overlook the dev env page delivery size because it doesn't have an impact on production? (I'm not saying it doesn't just is there a way that it can have no impact or is this only achieveable by using a cdn or static site generation?)

scottaubrey:
> > Nice catch on the bug.
> > Since we've resisted introducing libraries into the rendering stack before due to it necessarily increasing payload size, I thought it worth mentioning that this seems to have added about 2kb to page delivery. In context, our page delivery size for 85111 imported in dev env is 552kb (or 550kb). Nothing noticable really in the grand schema of things.
> 
> We should discuss the approach of resisting introducing libraries. Is there a way to minimise or offset the use of libraries. It would be great even for small utility functions to support light libraries. It obviously would be very easy to write this function but I believed I was doing the right thing by using a library. Are we able to overlook the dev env page delivery size because it doesn't have an impact on production? (I'm not saying it doesn't just is there a way that it can have no impact or is this only achieveable by using a cdn or static site generation?)

I was measuring the prod page size (using the `docker-compose.prod.yaml` which runs the build and serve rather than dev). The nature of our setup currently means that the react component is now dependant on the js library to render, so when it's shipped to client to allow rerender, so is the library. There isn't a way around that except to remove react hydration client-side, which would have a knock on effect to a few bits of functionality. Happy to chat it through though. My comment wasn't advocating one way or the other, more just preempting a potential question around the libraries inclusion as that has previously been resisted (for example in the axios vs fetch).

PR #2167, opened 2024-10-10
Introduce I18nextProvider

nlisgo:


nlisgo:
I've moved `i18n` from the components because it was only needed for the tests and storybook. To my mind, that means that is should be instead introduced in the tests and storybook.

Issue #2238, opened 2025-05-07
Fix sass warnings emitted when building Storybook


davidcmoulton:
This ticket is to fix on `master` as many of the sass compilation deprecation warnings as practicable. The classes of warning and examples, references and specimen fixes may be found in https://github.com/elifesciences/enhanced-preprints-client/issues/2237.

Note that once the `mixed declaration` warnings are addressed, try to add a sass linting rule to prevent any more being allowed through in the future.

**Update:** Added linting rule to help catch future problems like this in [977f7eb](https://github.com/elifesciences/enhanced-preprints-client/commit/977f7eb9cf818da9107cd958a11cdd0ea08d8586).

davidcmoulton:
This ticket's commits from various branches have now all been cherry-picked onto `master`. Closing this ticket as done.

PR #2239, opened 2025-05-20
[DO NOT MERGE] Fix some'mixed declaration' deprecation warnings during sass compilation 

davidcmoulton:
See #2238 for description.

This is being done on a branch rather than on `master` so I can raise a PR to trigger the visual regression checks without breaking the build when they fail, (which during this process they might).

Suggests missing tooling for local visual regression checks while developing?

### DIFF
--- a/src/components/atoms/article-flag/article-flag.scss
+++ b/src/components/atoms/article-flag/article-flag.scss
@@ -3,13 +3,13 @@
 @use "../../../sass/typography";
 
 .article-flag__link {
-  @include typography.msa-link-text();
-  @include mixins.padding(6 0);
-
   border-radius: 3px;
   cursor: pointer;
   display: inline-block;
   text-decoration: none;
+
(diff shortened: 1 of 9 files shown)


davidcmoulton:
[3dfdf57](https://github.com/elifesciences/enhanced-preprints-client/pull/2239/commits/3dfdf57f88d272850bd28c472c788ef82966babc) concerns the `padding` mixin: it wraps each emitted style rule in `& {}` to make it explicit these style rules are not intended to be moved up. This in turn has introduced a visual regression in the site footer for the horizontal padding of `site-footer__container`, suggesting there was a reliance on sass moving style rules up during compilation in order to get the desired result.

davidcmoulton:
The contents of all commits on the branch `address-sass-mixed-declarations` have now made it into `master`, independently of this PR. Closing this PR as redundant.

PR #2242, opened 2025-05-23
[DO NOT MERGE] Reorder style rules to avoid mixed declaration warnings during sass compilation

davidcmoulton:


davidcmoulton:
bfbbe2cd7417edc9a141f8a62cd3401e5d87ea9e uses the `stylelint-order` plugin to enforce an unambiguous order of style rules that avoid triggering the sass compiler's `mixed-declaration` deprecation warnings.

davidcmoulton:
Changed my mind. Decided to keep this change local so that the linting won't fail in CI
and so the chromatic tests will actually run on the PR.

davidcmoulton:
As [c85a37e](https://github.com/elifesciences/enhanced-preprints-client/pull/2242/commits/c85a37ea2123d78e92a767b492f3ade8851adf0f) shows, fixing the mixins before the style blocks that use them can cause visual regresssions as legitimate style overrides are harder to spot. 'Outside in' seems a better approach, so will revert the mixin changes and proceed fixing the selectors' style blocks before those of mixins.

